Issue #6, opened 2022-07-18
Training in a different window size than initially trained gives error


hessil:
I successfully trained a JABS classifier with a mix of v5 and v3 pose on a window size of 5. When I try a different window size (tried with 10 and 15) I get this error:

```
Traceback (most recent call last):
  File "/Users/hessil/Downloads/JABS-behavior-classifier-main 2/src/feature_extraction/features.py", line 333, in get_window_features
    features = self.__load_window_features(window_size)
  File "/Users/hessil/Downloads/JABS-behavior-classifier-main 2/src/feature_extraction/features.py", line 265, in __load_window_features
    with h5py.File(path, 'r') as features_h5:
  File "/opt/anaconda3/envs/myenv/lib/python3.9/site-packages/h5py/_hl/files.py", line 424, in __init__
    fid = make_fid(name, mode, userblock_size,
  File "/opt/anaconda3/envs/myenv/lib/python3.9/site-packages/h5py/_hl/files.py", line 190, in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 96, in h5py.h5f.open
OSError: Unable to open file (unable to open file: name = '/Users/hessil/Desktop/socialvideos/rotta/features/NV1-B2B+2019-07-19_AM+19062002_2019-07-20_12-00-00_54000/0/window_features_15.h5', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/hessil/Downloads/JABS-behavior-classifier-main 2/src/ui/training_thread.py", line 54, in run
    features, group_mapping = self._project.get_labeled_features(
  File "/Users/hessil/Downloads/JABS-behavior-classifier-main 2/src/project/project.py", line 671, in get_labeled_features
    window_features = features.get_window_features(
  File "/Users/hessil/Downloads/JABS-behavior-classifier-main 2/src/feature_extraction/features.py", line 339, in get_window_features
    features = self.__compute_window_features(window_size)
  File "/Users/hessil/Downloads/JABS-behavior-classifier-main 2/src/feature_extraction/features.py", line 468, in __compute_window_features
[...]

PR #7, opened 2022-08-20
implement fix for #issue6

gbeane:
Fix for issue #6

### DIFF
--- a/src/feature_extraction/feature_group_base_class.py
+++ b/src/feature_extraction/feature_group_base_class.py
@@ -53,7 +53,7 @@ class FeatureGroup(abc.ABC):
         feature_modules = self._init_feature_mods(identity)
         return {
             name: mod.window(identity, window_size, per_frame_values[name]) for name, mod in
-            feature_modules.items()
+            feature_modules.items() if name in per_frame_values.keys()
         }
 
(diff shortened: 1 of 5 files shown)


src/feature_extraction/feature_group_base_class.py
```diff
@@ -53,7 +53,7 @@ def window(self, identity: int, window_size: int,
         feature_modules = self._init_feature_mods(identity)
         return {
             name: mod.window(identity, window_size, per_frame_values[name]) for name, mod in
-            feature_modules.items()
+            feature_modules.items() if name in per_frame_values.keys()
```
gbeane, line 56: this line is the main fix

gerring:
It would appear I am not on the project and cannot review merge requests?

gerring:
Change looks okay but have little idea what it does or if it breaks or fixes things because lack of tests. I assume it is fine.

SkepticRaven:
I tested on one of my pose_v2 projects, and seems to re-calculate window features.
We'll see what Leinani's tests return, since it may be related to a specific combination of pose file versions available.

SkepticRaven (approved):

gbeane:
found one more thing I need to troubleshoot related to Leinani's issue

PR #8, opened 2022-08-24
Issue6

gbeane:
more correct fix for issue #6 

this ensures all features are loaded from previously computed hdf5 file. This way, when recomputing window features, we have all necessary features to compute all of the window features supported by a given pose file (even if the current project doesn't use all of those features)

### DIFF
--- a/src/feature_extraction/feature_group_base_class.py
+++ b/src/feature_extraction/feature_group_base_class.py
@@ -53,7 +53,7 @@ class FeatureGroup(abc.ABC):
         feature_modules = self._init_feature_mods(identity)
         return {
             name: mod.window(identity, window_size, per_frame_values[name]) for name, mod in
-            feature_modules.items()
+            feature_modules.items() if name in per_frame_values.keys()
         }
 
     def feature_names(self, features: typing.Optional[str] = None):
--- a/src/feature_extraction/feature_group_base_class.py
+++ b/src/feature_extraction/feature_group_base_class.py
@@ -53,7 +53,7 @@ class FeatureGroup(abc.ABC):
         feature_modules = self._init_feature_mods(identity)
         return {
             name: mod.window(identity, window_size, per_frame_values[name]) for name, mod in
-            feature_modules.items() if name in per_frame_values.keys()
+            feature_modules.items()
         }
 
     def feature_names(self, features: typing.Optional[str] = None):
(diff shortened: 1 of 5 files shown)


gbeane:
sorry @gerring, there are no tests added by this PR. This is about the max amount of effort that I can spend on this. Maybe a good intern project to mock everything up needed to add good unit test coverage for every class...

gbeane:
last PR fixed the exception Leinani was seeing adding a new window size via the GUI, but the newly generated window feature files only included features enabled in the current project. When adding additional window sizes using the command line script, the resulting window feature files contained all features supported for a given pose files, even if some of those features are unused by the current project. This change now matches that behavior when adding new window sizes in the GUI.

hessil:
Issue is still occurring for me

gbeane:
> Issue is still occurring for me

Did you checkout this branch?

hessil:
Yes

hessil:
I've just deleted the cache and features for the project and am running that, this will take a while to check

gbeane:
> I've just deleted the cache and features for the project and am running that, this will take a while to check

okay.  I was able to reproduce your exact error with the project you sent me with Dropbox, and this branch resolved it for me.

hessil:
Clearing cache+features seemed to do the trick! Haven't got an error yet training on a second window size

SkepticRaven (approved):

SkepticRaven:
Appears as though the test from the prior pull request broke the cache/feature folders. We should add "verify feature cache" as a feature request so that updating between versions goes a bit more smoothly.

gbeane:
> Appears as though the test from the prior pull request broke the cache/feature folders. We should add "verify feature cache" as a feature request so that updating between versions goes a bit more smoothly.

so feature h5 files and other cache files include a version number that gets bumped when an incompatible change in file format is introduced so they are automatically re-created if they're an out of date format.

Unfortunately in this case a bug in the previous pull request (that was closed without merging) appears to have caused a problem with cached files.

SkepticRaven:
> > Appears as though the test from the prior pull request broke the cache/feature folders. We should add "verify feature cache" as a feature request so that updating between versions goes a bit more smoothly.
> 
> so feature h5 files and other cache files include a version number that gets bumped when an incompatible change in file format is introduced so they are automatically re-created if they're an out of date format.
> 
> Unfortunately in this case a bug in the previous pull request (that was closed without merging) appears to have caused a problem with cached files.

Makes  sense. Probably not necessary then if its limited to potential changes breaking stuff.

PR #12, opened 2023-04-10
Tc 245

SkepticRaven:
Pull for integrating all the segmentation features and logging challenges related to it.


SkepticRaven:
Known issues at this timepoint

1. Poses are not currently matching segmentation.
seg-mismatch

2. Multiple contours per mouse are flattened to 1 contour per mouse (instead of a list of contours).
multi-contour-issue

3. Pixels to cm conversion isn't being applied to moments correctly (m00 is area, and needs to be multiplied by conversion^2)
`contour_moments = [[cv2.moments(trim_seg(np.reshape(seg_data[0,x,y], [-1,2])))['m00']*(cm_per_px**2)*[-1,1][seg_external_flag[0,x,y].astype(int)] for y in np.arange(np.shape(seg_data)[2])] for x in np.arange(np.shape(seg_data)[1])]`
Instead of figuring out how to merge the moments, it may also be simpler to render the frame and just pass that to `cv2.moments`

4. My first edit for enabling the features to get passed in causes issues with other pose formats -- `AttributeError: 'PoseEstimationV5' object has no attribute 'get_seg_id'` is thrown when I have a project with only pose_v5 files.

SkepticRaven:
This pull is being closed in favor of https://github.com/KumarLabJax/JABS-behavior-classifier/pull/16

PR #13, opened 2023-04-24
Apple silicon compatibility

anshu957:
`Description:`
This pull request adds support for Mac M1/M2 devices by setting the QT_MAC_WANTS_LAYER environment variable to 1. This is required to ensure proper rendering of the application's graphical interface.

`Changes:`
- Added the following code to the app.py:
```python
    #### For Mac M1 compatibility ######
    import os
    os.environ['QT_MAC_WANTS_LAYER'] = '1'
    ####################################
```

- Added conda environment yaml file for users who want to install using conda as pip has problems installing `pyside2` on Mac M1/M2 (https://forum.qt.io/topic/124289/mac-m1-pyside-2-6-can-t-install-lib/8) 
    
`Testing:`
Tested this on 2 different mac M1 devices and verified that the application's graphical interface is rendered correctly.


SkepticRaven:
Although it was linked to the issue, it was not mentioned in this pull request:

Going from shapely 1.7 -> 2.0 seems to change the way of accessing centroid keypoints.
This patch should fix it: https://github.com/KumarLabJax/JABS-behavior-classifier/pull/27/commits/188fe7a654d55a4a34b2c689f1f8fe457ea66929

environment_jabs.yml
SkepticRaven, line 1: Before a merge, we should probably add installation instructions w/ conda, since this install deviates from current instructions.

All these libraries are also drastically different versions from the ones in requirements.txt
As mentioned elsewhere PyQT5/pyside2 are the primary reasons for why conda was used here. It looks like pyside2 is no longer getting packport support for newer hardware (in place of pyside6). It doesn't look like that upgrade would be too difficult: https://www.pythonguis.com/faq/pyside2-vs-pyside6/

SkepticRaven (changes requested):
Required changes:
* Test patch from https://github.com/KumarLabJax/JABS-behavior-classifier/pull/27/commits/188fe7a654d55a4a34b2c689f1f8fe457ea66929 for multimouse project feature generation crash
* Add docs for conda installation

Alternative Change:
* Check compatibility with pyside6 and use that instead of a conda env.

SkepticRaven (approved):

SkepticRaven:
Added the 2 patches and bumped the version.
I didn't run extensive tests, but I did run `initialize_project.py` on a multimouse project (via the venv libraries) to make sure the patches were compatible.

If @anshu957 wanted to do one last check for this, you can... otherwise I think we can merge/release.

anshu957:
Thanks @SkepticRaven. I just validated it for a single mouse dataset and I could train and export with all three algorithms without any problem. It is displaying the overlaying poses and tracks correctly as well. I think we can merge it and release it.

PR #14, opened 2023-05-11
Bug1 - Segmentation Data Sorting

nick-sebasco:
I have addressed the first bug.  The segmentation data now aligns with the pose data,
mouse_segmentation_working
For example (see above image), previously frame 950 was a case where the segmentation data failed to align with mouse 0, now it always matches the pose data.


src/video_stream/frame_annotation.py
```diff
@@ -212,97 +212,97 @@ def overlay_pose(img: np.ndarray, points: np.ndarray, mask: np.ndarray,
...
-    # Discard values < 0.
-    contour = contour[contour >= 0]
-    if len(contour) > 0: 
-        # Filter missing contours and coerce type to satisfy drawContours
-        # method.
-        contour = contour.reshape((len(contour)//2, 2)).astype(int)
-
-        cv2.drawContours(img, [contour], -1, _ACTIVE_COLOR, 2)
-
+def overlay_segmentation_DEPRECTAED(img: np.ndarray, pose_est: PoseEstimation,
```
SkepticRaven, line 215: Are these functions used or were they just for prototyping? It may be a good idea to leave them to the git history and delete them.

SkepticRaven:
Functionality is working as intended in my hands!

PR #18, opened 2023-05-15
Tc 272 - Use Moments Rather Than Convex Hull for Label

nick-sebasco:
Another item from the backlog.  Now the primary means of computing the label position is the image moments.  If the label position computed from moments is not a valid number, the old convex hull method is used.

**Before (Frame 3108)**
label_before

**After (Frame 3108)**
label_after


SkepticRaven:
Closing this because the branch is a bit too stale to try and resolve merge conflicts. When we get back to this feature, we should be able re-use the implementation strategy.

PR #23, opened 2023-07-11
Label balancing and symmetric behaviors for training classifiers

SkepticRaven:
Adds:
* label downsampling routine for training classifiers with balanced labels
* GUI checkbox for optionally conducting the balanced training
* balanced training value stored in both trained classifier + project
* random subsetting is hit by random_seed
* bugfix on storing selected gui option for social features


SkepticRaven:
Current TODO:
* Update all calls of `classifier.train` to now include new `uses_balance` parameter

Right now, this option only really works within the gui. classify.py will fail.

src/project/export_training.py
```diff
@@ -39,8 +42,11 @@ def export_training_data(project: 'Project',
     writes exported data to the project directory
     :param project: Project from which to export training data
     :param behavior: Behavior to export
+    :param pose_version: Minimum required pose version for this classifier
```
dahhei, line 45: Are all pose versions currently backwards-compatible?

SkepticRaven: Features are backwards compatible, but projects/classifiers can only operate on the lowest common feature set. This is the current way projects calculate this number.

Currently no logic is present for decreasing the number passed here... but it could be added (rather than just relying on param1's value).

dahhei:
I believe that I can't test these changes on my local install of JABS unless the update-reqs is merged first. There is a version from commit-head 49c9b10 on the HPC that I will use to test these updates. 😃

dahhei:
Cayson initially raised this issue so I tested out the all k-fold cross-validation and it appears as though the top 10 feature labels are incorrect. I trained a Drinking classifier which typically has the distance to Lixit features present. This one does not. The all k-fold does seem to work otherwise though.

dahhei:
It would be nice if the cross validation iteration value included a "cross validation iteration 14/n", with n being the total number that it will test.

SkepticRaven:
The commits above are a rebase merge with main (based on the other recent merge into main: https://github.com/KumarLabJax/JABS-behavior-classifier/pull/27)

SkepticRaven:
The feature rewrite to propagate names alongside values fixed the pose_v2 bug.
This branch should be ready to merge (after updating the appropriate version numbers)!

Issue #24, opened 2023-07-19
xgboost doesn't use multiple threads for training / upgrading library versions


SkepticRaven:
The settings are set correctly, but it appears to be a bug with the library version we're using: https://discuss.xgboost.ai/t/n-jobs-1-no-longer-uses-all-cores/1955/4

A quick test upgrading from 1.3.1 to 1.7.1 works fine in the singularity image.
There should be a dev branch created to upgrade the 3+ year old frozen libraries and ensure everything still works as intended.

Issue #5, opened 2022-06-06
App not opening in M1 Mac -- PyQt possible culprit


anshu957:
I'm working with M1 Mac -- Monterey 12.4 with  PyQt5.15 and when running `python3 app.py` the window freezes indefinitely. Upon searching, the only solution that worked for me is setting environment variable `export QT_MAC_WANTS_LAYER=1`. 
Is there a better way to fix this?

gbeane:
We could probably set that env variable in JABS _**before**_ importing PySide2 (We don't use PyQt5.x, we are using PySide2 for our Python Qt bindings).

I'd try making this change to `app.py`:

```python
import argparse
import sys
import os

os.environ['QT_MAC_WANTS_LAYER'] = '1'
from PySide2 import QtWidgets

from src import APP_NAME, APP_NAME_LONG
from src.ui import MainWindow
```

if you could test that and let me know (make sure you're not also setting it in your shell) that would be great (I don't have access to a M1 mac to test on)

anshu957:
Thanks! Yes, it is working without exporting the variable in the shell.

gbeane:
okay, thanks for testing.

I will make a pull request with this change.

SkepticRaven:
Adding recent notes to this:

Anshul created a pull request using a conda environment, since a bunch of the GUI libraries aren't on pypi yet as pre-compiled binaries: https://github.com/KumarLabJax/JABS-behavior-classifier/pull/13

Single mouse appears to operate just fine in that conda env, but it appears that feature calculation for multi-mouse isn't operating correctly. User log from trying to to train a multimouse classifier (upon train button click):

```
Traceback (most recent call last):
  File "/Users/hamilc/Jax/JABS-behavior-classifier/src/feature_extraction/features.py", line 113, in __init__
    self.__load_from_file()
  File "/Users/hamilc/Jax/JABS-behavior-classifier/src/feature_extraction/features.py", line 152, in __load_from_file
    with h5py.File(path, 'r') as features_h5:
  File "/Users/hamilc/anaconda3/envs/jabs/lib/python3.10/site-packages/h5py/_hl/files.py", line 533, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
  File "/Users/hamilc/anaconda3/envs/jabs/lib/python3.10/site-packages/h5py/_hl/files.py", line 226, in make_fid
[...]

PR #27, opened 2023-07-20
Update library requirements.txt

SkepticRaven:
Updating the requirements.txt to not be 2-3 year old libraries.
There were a couple other changes to get compatibility working again.

Tests run at the beginning of this pull:
* Ran unittests: `python3 -m unittest discover -s /behavior-classifier/tests/`
  * `Ran 33 tests in 0.785s   OK`
* Ran `python3 initialize_project.py` on a project
* Opened the GUI from an updated gui-vm and trained a classifier

I don't have a bunch of dev environments to test, so these were all conducted inside singularity (running debian).
One of the fixed I needed to do (upgrade to shapely change) may address https://github.com/KumarLabJax/JABS-behavior-classifier/issues/5#issuecomment-1621725227
Also address https://github.com/KumarLabJax/JABS-behavior-classifier/issues/24

Would be good to have someone with a mac environment to share their experience.


SkepticRaven:
Works successfully on windows 10 (python 3.10.11)

SkepticRaven:
@dahhei tried to install the venv on a mac and discovered shiboken needed to be updated from 5 -> 6

SkepticRaven:
This branch would have particularly benefited from having more coverage on unittests.
Given that I missed that one, I should probably click all the buttons on the gui and even run tests on the non-gui version to make sure things don't change before a merge.
New list of tests that should probably go through:
* Click all the buttons on gui
* Test multiple project pose versions (2, 3, 4, 5)
* Test command line interface (classify.py, initialize_project.py, and stats.py)

Also I upgraded the singularity gui vm from buster -> bookworm, so should make the non-gui one also be bookworm (not bullseye).

dahhei (approved):
The requirements.txt works without errors on my local M1 Macbook. Exporting a classifier also works fine on my end now.

SkepticRaven:
A couple notes from testing/poking around:
* (Critical) Pause (after play) causes segfault ~50% of the time
  * Currently unclear cause, as threads seem to get closed correctly. It's happening after/during the repaint even after it stops
* (Minor) Text littered around the UI is now lighter compared to older versions
  * This may be linked to OS settings and may be limited to my Ubuntu host machine (not present in CentOS test)

SkepticRaven:
Latest commit should clear the 2 issues above.
Remaining checks before merge:
* Confirm crash fix native on macs
* Test `stats.py`

PR #36, opened 2024-01-16
Select feature subsets

SkepticRaven:
Adds the ability to control feature subsets for classifiers.
Changes ownership of project settings from UI components to project.
Classifiers make a copy of project settings.


src/ui/central_widget.py
```diff
@@ -302,13 +282,14 @@ def overlay_segmentation(self, new_val: bool):
     def remove_behavior(self, behavior: str):
         self._controls.remove_behavior(behavior)
 
-    def _change_behavior(self):
+    def _change_behavior(self, new_behavior):
```
SkepticRaven, line 285: While this isn't used in this function, it is added here to allow the new behavior string to be available from the signal that calls this. This variable is currently owned by the controls object and can be accessed via `self.behavior`.

PR #41, opened 2024-07-29
Update prediction format

SkepticRaven:
Changes:
* Adjusted the prediction format
  * Prediction file now contains behavior key instead of relying on implied folder structure
  * Multiple predictions can now be placed into the same prediction file
* Adjusted the internal representation for feature cache to match updated classifier format (feature-named column vectors)
* Added new logging attributes for predictions (e.g. pose hash, classifier used)

While digging around in the code, I also patched and extended some items (scope creep in this branch):
* Patched bugs related to incorrect 0-paddings -> NaNs
* Patched bug related to min/max features
* Added new arena corner features
* Added new lixit features


SkepticRaven:
@dahhei For this review, I'm only looking for review on doc changes. This pull request has suffered too much from scope creep to reasonably ask you to review the 1k code line changes.
I also recently discovered a new bug with features, but want to get this pull request through before adding even more scope creep. If the docs look in line, I'll verify that the relevant versions get bumped and merge.

dahhei:
The documentation looks great! I like the strings that clearly specify the counts for each feature type. Seems like you changed the appropriate docs with your edits and also made some of the docstrings more detailed. No concerns from my end

PR #43, opened 2024-09-09
Patching keypoint masking and cm units

SkepticRaven:
Bug patches:
* Some pose masking did not properly threshold confidences
* Masking of keypoints not tiled correctly to x,y locations
* Pixel units never applied properly to some features
* FPS feature not applied to some segmentation features
* Adjusted angle features to be "mathematicians" angle (0 heading forward, + counter-clockwise)

Internal Change
* To avoid future confusion, all data is now represented internally as (x, y). Data is transformed at pose file read time. Cached data is always (x,y). Handful of incorrect comments were corrected/removed. Features were adjusted to remain correct.

### DIFF
--- a/src/pose_estimation/pose_est_v5.py
+++ b/src/pose_estimation/pose_est_v5.py
@@ -1,10 +1,14 @@
 import typing
 from pathlib import Path
-
+import numpy as np
 import h5py
 
 from .pose_est_v4 import PoseEstimationV4
 
+OBJECTS_STORED_YX = [
+    'lixit',
+    'food_hopper',
+]
 
 class PoseEstimationV5(PoseEstimationV4):
     def __init__(self, file_path: Path,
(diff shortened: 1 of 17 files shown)


src/pose_estimation/pose_est_v5.py
```diff
@@ -1,10 +1,14 @@
 import typing
 from pathlib import Path
-
+import numpy as np
 import h5py
 
 from .pose_est_v4 import PoseEstimationV4
 
+OBJECTS_STORED_YX = [
```
dahhei, line 8: good note to have

dahhei:
Most of what I saw was you swapping around x and y values when necessary, and adjusting the masking functions to work with a new minimum_confidence. This all looked good to me, documentation also looks good (doesn't look like you really added new functions that would need new docstrings). I didn't test any of the actual functionality, I can try if you want me to, but I assume we'd find out if anything errors pretty quickly and I'd probably not be a great tester for each individual change. 

All in all, I have no issues and think this can be merged.

dahhei (approved):

PR #44, opened 2024-10-03
OFA features

SkepticRaven:
Adding some features that get calculated and cached but not used in classifiers. These features are designed to enable traditional open field metrics to be calculated, as well as enabling more granular information about spatial location to be available for behavior calls.

Also found bugs with "nearest wall" feature, so still needed to bump the version

Exact Changes
* Added "distance to all walls" to be stored in feature cache file
* Added "average wall distance" to be stored in feature cache file
* Fixed a bug where mouse centroid would get px -> cm conversion twice in wall distance calculation
* Fixed a bug where wall corner points were assumed to be sorted, when they are not guaranteed to be sorted... by sorting them

### DIFF
--- a/docs/features/features.md
+++ b/docs/features/features.md
@@ -157,11 +157,20 @@ Per-frame features fill missing values with zeros before passing into the FFT.
 
 # Extra Features calculated, but not used in a classifier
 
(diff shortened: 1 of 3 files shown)


SkepticRaven:
Adding a note here: this pull is currently waiting on validation experiments (e.g. that wall distance can actually be used for periphery time).
The associated pull is [over here](https://github.com/KumarLabJax/JABS-postprocess/pull/6). The associated repo is currently private.

SkepticRaven:
Experiments resulted in passing validation experiments.

PR #46, opened 2024-10-15
Add working `environment.yml`

talmo:
This PR adds an `environment.yml` that produces a working conda environment across platforms and corresponding instructions in the README.

The only non-conda package that is used is for `PySide6` since `QtWebEngine` is not yet available for Qt 6.x on `conda-forge` (https://github.com/conda-forge/qt-webengine-feedstock/pull/46).

I didn't add anything to the docs in case you guys wanted to hold off on that before testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ estimation neural networks. Single mouse pose files are generated from [this rep
 ## Requirements
 Developed and tested on Python 3.10. See the `requirements.txt` 
 for a list of required Python packages. These packages are available from the 
-Python Package Index (PyPI)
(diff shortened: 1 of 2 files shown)


anshu957:
Tested on M1 chip with python version --3.10.15:
- Opening GUI app
- training the classifier
- labeling the frames 

All OK ✅

SkepticRaven (approved):
There's some issues with getting libxcb to link up for the linux conda environment, but I think that's more of a headless or system/conda path issue which we can ignore for now...

PR #47, opened 2024-10-21
Patch 0.17.3 tests

SkepticRaven:
Changes 2 test assertions to allow nans in comparison.


dahhei (approved):
Changes look good to me, I'm assuming it passes all tests now. I can go back and actually test it myself if you want another confirmation that the bug is patched but it looks fine to me.

SkepticRaven:
Correct, these patches actually came after I attempted to re-build the container (which runs the tests during build) after the 0.17.3 release. Should probably look into seeing what costs may be associated with some sort of automated testing during merges...

PR #53, opened 2024-12-20
Packaging feedback

SkepticRaven:
Changes:
* Updating container definitions to use poetry
  * builds wheel
  * installs wheel
  * cleans up
* Removing 2 warnings from newer lib versions
  * numpy arr into scalar (was arr of length 1)
  * scipy warning for nan values in circular stats calls
* fixing some segmentation tests
  * these were never adequately checked because they were missed by the "discover". They still need some work to actually test well, but this is good enough for now...

### DIFF
--- a/src/jabs/feature_extraction/feature_base_class.py
+++ b/src/jabs/feature_extraction/feature_base_class.py
@@ -2,6 +2,7 @@ import abc
 import typing
 import numpy as np
 from scipy import signal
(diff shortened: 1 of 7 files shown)


gbeane (approved):
Approved. Minor comment in the singularity .def file. Not sure if there is anything that needs to be addressed or not, but it seems to be functional as is.

PR #54, opened 2025-01-01
Update packaging to use poetry

gbeane:
Uses Poetry for managing dependencies. 
Makes JABS pip installable from git repo
allows building Python wheel with Poetry
installs four commands into Python environment:
jabs (launches JABS GUI)
jabs-init (replaces initialize_project.py)
jabs-classify (replaces classify.py)
jabs-stats (replaces stats.py)


SkepticRaven (approved):
We had discussed most items over email prior to this request.

The only major note that I'm adding from that thread is that we're keeping leaving the singularity containers to use poetry rather than pip for now to allow it to look at poetry.lock for better versioning consistency.

PR #55, opened 2025-02-04
Standalone generate features

SkepticRaven:
Adds `jabs-features`, which calculates a feature cache for a single pose file.
This largely overlaps with `jabs-init`, but enables some new workflows.

Newly enabled characteristics:
* Feature cache generation no longer needs to be attached to either a project or classifier.
* Calculated features can be requested at a potential downgrade by user (e.g. force pixel units)

Additionally fixed some docs to refer to the `jabs-*` instead of `*.py`.


dahhei (approved):
Looks good to me. Did not test, please let me know if you want me to.

Docs look good, the changes make sense to me.

SkepticRaven:
Minor note, I'm testing this a bit more thoroughly while working on a nextflow pipeline.

It appears that there may be a bug related to `jabs-classify` when it tries to use cached features with cm-units...

Classifier suggests: `CM Units: 1`
However, `File "/usr/local/lib/python3.10/site-packages/jabs/feature_extraction/features.py", line 192, in __load_from_file` throws `DistanceScaleException`.
Inspecting the feature cache and pose file, the distance scale factor is equal. Going to raise a new issue and maybe attach a cache checking patch to this pull (if it's relatively trivial).

SkepticRaven:
> It appears that there may be a bug related to `jabs-classify` when it tries to use cached features with cm-units...

Correction: This is actually working at intended. The nextflow was missing `--use-cm-distances`, so the features were in pixels and it changed the attributes before failing later (trying to overwrite a field).

We should still try and make it not fail like that, so adding a slightly different ticket as an enhancement. I forgot how updating feature cache is handled in the UI, but right now the code can't overwrite an old cache.

gbeane (approved):

PR #57, opened 2025-02-24
Add eccentricity calculations

luciazhxng:
Note: I manually chose a kernel for erosion/dilation, which probably isn't the smartest but it worked for the videos I think

### DIFF
--- a/src/jabs/feature_extraction/segmentation_features/moment_cache.py
+++ b/src/jabs/feature_extraction/segmentation_features/moment_cache.py
@@ -1,6 +1,6 @@
 import numpy as np
 import cv2
-
+import matplotlib.pyplot as plt
 from jabs.pose_estimation import PoseEstimation
 
 
@@ -25,6 +25,8 @@ class MomentInfo:
         self._moments = np.zeros((self._poses.num_frames, len(self._moment_keys)), dtype=np.float32)
         self._seg_data = self._poses.get_segmentation_data(identity)
         self._seg_flags = self._poses.get_segmentation_flags(identity)
+        self._contours = [None] * self._poses._num_frames
+
 
         # Parse out the contour matrix into a list of contour lists
         tmp_contour_data = []
@@ -36,8 +38,10 @@ class MomentInfo:
         for frame, contours in enumerate(self._seg_data):
             # No segmentation data was present, skip calculating moments
             if len(contours) < 1:
+                
                 moments = {key: np.nan for key in self._moment_keys}
             else:
+                self._contours[frame] = self.calculate_contours(contours)
                 moments = self.calculate_moments(contours)
             # Update the output array with the desired moments for each frame.
             for j in range(len(self._moment_keys)):
@@ -62,6 +66,9 @@ class MomentInfo:
         """
         key_idx = self._moment_keys.index(key)
         return self._moments[frame, key_idx]
+    
+    def get_contours(self, frame):
+        return self._contours[frame]
 
     def get_all_moments(self, frame):
         """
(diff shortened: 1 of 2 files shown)


src/jabs/feature_extraction/segmentation_features/moment_cache.py
```diff
@@ -119,3 +126,57 @@ def calculate_moments(contour_list):
...
+        """
+        frame_size = [800, 800]
+        mask = np.zeros(frame_size, dtype=np.uint8)
+        cv2.drawContours(mask, contour_list, -1, (1), thickness=cv2.FILLED)
+        kernel = np.ones((7, 7), np.uint8)
+        eroded_mask = cv2.erode(mask, kernel, iterations=1)
+        dilated_mask = cv2.dilate(eroded_mask, kernel, iterations=1)
+        contours, _ = cv2.findContours(dilated_mask, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_SIMPLE)
+
+        # if len(contours) > 0:
```
gautam-sabnis, line 147: Please remove the commented-out code if it's not going to be part of the final version.

src/jabs/feature_extraction/segmentation_features/moment_cache.py
```diff
@@ -119,3 +126,57 @@ def calculate_moments(contour_list):
...
+        :param img: The current video frame.
+        :param seg_data: This will be the segmentation for a particular frame and indentity.
+        :param color: color of segmentation contours rendered on the GUI.
+        :return: None
+        """
+        frame_size = [800, 800]
+        mask = np.zeros(frame_size, dtype=np.uint8)
+        cv2.drawContours(mask, contour_list, -1, (1), thickness=cv2.FILLED)
+        kernel = np.ones((7, 7), np.uint8)
+        eroded_mask = cv2.erode(mask, kernel, iterations=1)
```
SkepticRaven, line 143: [Future work]
Erosion/dilation parameters should likely be somehow attached to pixel information.
Here, Lucia uses a 7x7 kernel with 1 iteration with the intent to remove the tail for her dataset (OFA, similar to https://doi.org/10.7910/DVN/SAPNJG). If we want to generalize this, we should probably get an average thickness of tail (in cm) and determine the kernel size via the number of pixel radius to adequately erode a feature of that size.

src/jabs/feature_extraction/segmentation_features/moment_cache.py
```diff
@@ -119,3 +126,57 @@ def calculate_moments(contour_list):
...
+    def calculate_contours(self, contour_list):
+        """
+        Draw all contours given data for a particular mouse in a particular video frame.
+
+        :param img: The current video frame.
+        :param seg_data: This will be the segmentation for a particular frame and indentity.
+        :param color: color of segmentation contours rendered on the GUI.
+        :return: None
+        """
+        frame_size = [800, 800]
```
SkepticRaven, line 139: [Future work]
Frame size is arbitrary and while 800x800 will work for this data, it should probably depend on max of contour data rather than hard-coded values. Would need to think a bit more, but may need a buffer edge with morphological filters (although erode should only shrink, should be fine...).

src/jabs/feature_extraction/segmentation_features/moment_cache.py
```diff
@@ -25,6 +25,8 @@ def __init__(self, poses: PoseEstimation, identity: int,
         self._moments = np.zeros((self._poses.num_frames, len(self._moment_keys)), dtype=np.float32)
         self._seg_data = self._poses.get_segmentation_data(identity)
         self._seg_flags = self._poses.get_segmentation_flags(identity)
+        self._contours = [None] * self._poses._num_frames
```
SkepticRaven, line 28: Name of attribute should better describe the content. These are morphologically filtered contours, as opposed to  "contours" (already present in `self._seg_data`)

src/jabs/feature_extraction/segmentation_features/moment_cache.py
```diff
@@ -62,6 +66,9 @@ def get_moment(self, frame, key):
         """
         key_idx = self._moment_keys.index(key)
         return self._moments[frame, key_idx]
+    
+    def get_contours(self, frame):
```
SkepticRaven, line 70: Missing docstring

src/jabs/feature_extraction/segmentation_features/shape_descriptors.py
```diff
@@ -86,6 +89,17 @@ def per_frame(self, identity: int) -> np.ndarray:
             euler_number[frame] = np.sum(contour_flags[:len(contour_list)]==1)-np.sum(contour_flags[:len(contour_list)]==0)
             hole_area_ratio[frame] = (hole_areas * self._pixel_scale**2)/self._moment_cache.get_moment(frame, 'm00')
 
+            # CALCULATE ECCENTRICITY
+            contours = self._moment_cache.get_contours(frame)
+            if len(contours) > 0:
```
SkepticRaven, line 94: How does this handle missing segmentation (e.g. the `None` initialization [here](https://github.com/KumarLabJax/JABS-behavior-classifier/pull/57/files#diff-9327ffdabb94234bb2ebc3edc44ac143717f60f165ce9028066f8253908659b3R28))?
You may also need to check if `contours is not None and ...`.

SkepticRaven (commented):
Looks like the code is likely doing what you want it to.
2 comments for future generalization.
3 comments on docstrings/better descriptions.
1 question about a potential bug/handling of missing data.

dahhei (commented):
Nice changes, seems like you cleaned up the commented out code + added missing docstrings from Brian and Gautam's comments. I second Brian's assessment on not merging this branch at this stage

keithshep:
This PR will be closed but the associated branch `eccentricity` will remain.

PR #60, opened 2025-03-25
update project.json when adding new behavior labels

### LINKED ISSUE #49: Project data sometimes not saved correctly
Symptom:
Sometimes in a project with multiple behaviors, a behavior will disappear from the menus.

Solution:
Until this is patched, you can simply add a new behavior with exactly the same name. Existing annotations should re-appear.

Details:
As suggested above, annotations are not affected, only project.json.
This was likely introduced at some point when I was modifying the project settings in the [v0.17.1](https://github.com/KumarLabJax/JABS-behavior-classifier/releases/tag/v0.17.1) tag.
I suspect the exact issue is a missing UI socket that should be triggering updates to the project.json file.
There may be other buttons that are not adequately modifying classifier states (other have reported feature subsetting flags not being correctly changed).

gbeane:
closes #49

### DIFF
--- a/src/jabs/project/project.py
+++ b/src/jabs/project/project.py
@@ -252,8 +252,8 @@ class Project:
         """
         get the project metadata and preferences.
 
-        Returns a copy of the metadata dict, so that self._info can't be
-        modified
+        Returns a copy of the metadata dict, so that self._metadata can't be
+        modified by the caller
         """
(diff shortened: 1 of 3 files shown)


SkepticRaven (approved):
Changes look good and appear to resolve the issue.

PR #62, opened 2025-03-25
61 archiving last behavior in a project should be handled better

### LINKED ISSUE #61: Archiving last behavior in a project should be handled better
Using the File-Archive Behavior menu action

if a user archives all behaviors in a project, the labeling buttons remain active.  Perhaps the behavior should be consistent with opening a project with no behavior labels defined: a dialog opens and forces the user to create a new behavior.

gbeane:
closes #61

archiving the last behavior in a project behaves consistently with opening a project with no behaviors defined: a dialog is shown for the user to create a new behavior

dialog has been updated to allow user to quit JABS instead of saving a new behavior label (for example, if they open the wrong directory by mistake when selecting a project directory in the file open dialog)

### DIFF
--- a/src/jabs/ui/archive_behavior_dialog.py
+++ b/src/jabs/ui/archive_behavior_dialog.py
@@ -46,10 +46,15 @@ class ArchiveBehaviorDialog(QDialog):
 
     def __archive(self):
         behavior = self._behavior_selection.currentText()
-        self.behavior_archived.emit(behavior)
 
         # remove from combo box
         self.__remove_behavior(behavior)
+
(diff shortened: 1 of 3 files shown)


SkepticRaven (approved):
Changes look good. Definitely like the clarity change of "cancel" to "quit".

gbeane:
> Changes look good. Definitely like the clarity change of "cancel" to "quit".

improved the behavior of the archive dialog, especially when archiving all behaviors

previously if you quit JABS instead of adding a new behavior label, it didn't save the project.json file so the list of behaviors still included the archived behavior

SkepticRaven (approved):

PR #63, opened 2025-04-03
move project loading into a separate QThread to keep main UI thread responsive

gbeane:
This change moves the loading of a project into a separate QThread

this prevents the "spinning beach ball" on macOS or the "this program is unresponsive" error on Windows

While the project is being loaded in its own thread, the main window will display an indeterminate progress bar animation 

If an exception is raised while loading a project, that Exception is passed back to the main UI thread and the string representation is displayed in an alert dialog 

also adds type hints in main_window.py

### DIFF
--- a/src/jabs/ui/main_window.py
+++ b/src/jabs/ui/main_window.py
@@ -2,13 +2,15 @@ import sys
 
 from PySide6 import QtWidgets, QtCore, QtGui
 from PySide6.QtCore import Qt
+from PySide6.QtGui import QKeyEvent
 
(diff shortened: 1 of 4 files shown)


src/jabs/ui/main_window.py
keithshep, line 1: All looks good to me. I like the addition of type annotations. Not directly related to this MR but it made me think maybe we want to add python logging at some point to log application lifecycle and major events?

keithshep (approved):

SkepticRaven (approved):

PR #64, opened 2025-04-04
fix video player crashes

### LINKED ISSUE #19: UI Slider Crashes App
Moving the UI slider to fast crashes the application.

https://github.com/KumarLabJax/JABS-behavior-classifier/assets/119430052/d80ee1f5-d6ac-41d6-85da-cc9a7d81ef19

Reflected in Jira TC-270.

gbeane:
closes #19 

simplifies thread lifecycle management in the video player widget Now, a single _PlayerThread is created when the first video is loaded and it lives until JABS is terminated.

This fixes crashes that could occur when moving slider during video playback or pausing/unpausing with newer versions of PySide6

also updates PySide6 dependency version and regenerated poetry.lock

**note:** I had some weirdness when upgrading PySide6 in an existing virtualenv (import errors)

I found other instances of this on Stackoverflow.  I ended up having to uninstall everything related to PySide6 (PySide6, PySide6_Addons, PySide6_Essentials, and shiboken6) and then run poetry install to install the newer version.  Just keep this in mind in case you have any issues if upgrading PySide6 in an existing environment.


keithshep (approved):
Didn't see any problems

PR #65, opened 2025-04-05
add file menu action to clear project cache

gbeane:
adds an action in the File menu to clear a project cache

opens a dialog for the user to confirm

forces reload of the currently selected video so the pose is loaded from the original .h5 file


keithshep:
When would a user clear the project cache?

keithshep (approved):
Looks good to me. If we add logging at some point the `except OSError` should emit something

gbeane:
> When would a user clear the project cache?

it should be infrequent, but we've had some incompatibilities in cached files between versions so they needed to be cleared and regenerated.  At some point I am going to add some version information into the cache so JABS can deal with this automatically.

gbeane:
@keithshep  I'm not convinced this is necessary 

I'm going to hold off on merging this. I think a better solution would be to add some versioning  and validation of any cached file in the project directory, and invalidate anything missing this versioning (so all cached files created by previous versions of JABS would be re-generated)

SkepticRaven:
It's hit us before (e.g. #4), which happened to be working with v5 files. I know there's a version for features and re-generating those, but it looks like the cache version numbers are managed per-inherited class. A cursory glance shows that it's only present for [v3](https://github.com/KumarLabJax/JABS-behavior-classifier/blob/main/src/jabs/pose_estimation/pose_est_v3.py#L20) and [v4](https://github.com/KumarLabJax/JABS-behavior-classifier/blob/main/src/jabs/pose_estimation/pose_est_v4.py#L19), so since most of our recent stuff is v6 it's not being checked/managed.

src/jabs/ui/main_window.py
```diff
@@ -421,6 +429,37 @@ def _project_load_error_callback(self, error: Exception):
...
+            QtWidgets.QMessageBox.No
+        )
+
+        # restore the original setting
+        app.setAttribute(Qt.ApplicationAttribute.AA_DontUseNativeDialogs, dont_use_native_dialogs)
+
+        if result == QtWidgets.QMessageBox.Yes:
+            self._project.clear_cache()
+            # need to reload the current video to force the pose file to reload
+            self._central_widget.load_video(self._project.video_path(self.video_list.selected_video))
```
bergsalex, line 460: What happens if `self._project.video_path(self.video_list.selected_video)` returns `None`?

src/jabs/ui/video_list_widget.py
```diff
@@ -54,6 +55,12 @@ def _selection_changed(self, current, previous):
         """ signal main window that use changed selected video """
         if current:
```
bergsalex, line 56: Does this need to have an `else` branch for when current doesn't evaluate to True? e.g. Should `self._selected_video` be set to `None`?

bergsalex (approved):

src/jabs/ui/main_window.py
gbeane: if a project is opened successfully, there is always a selected video 
the clear cache menu is disabled unless a project has been loaded, so this shouldn't happen, but I'll add some more defensive programming

PR #69, opened 2025-04-07
support Python 3.10 - 3.12

gbeane:
updates packaging to support additional versions of Python


gbeane:
ignore the long list of commits -- some of those were merged already, but since they were squashed git doesn't think those commits are in main.  If you look at the actual changes being merged in on the Files Changed tab you can see there are only a few changes:

this mostly changes the pyproject.toml file to make some minor adjustments to package versions so that JABS can install in python 3.10 - 3.12

also renames a variable to a more suitable name in player_widget.py

keithshep (approved):
looks good

PR #70, opened 2025-04-07
Update setup_windows.bat to also support python 3.11 and 3.12

gbeane:
the setup_windows.bat script was only checking for python 3.10

update to also check for 3.11 and 3.12

also add a command line parameter to skip the python version check and just try to install


keithshep (approved):
🤞

PR #80, opened 2025-04-11
Update sklearn dependency

gbeane:
updates sklearn dependency to be >=1.5.0, <2.0.0

the 1.3.X series has a known bug (previously we were pinned to 1.3.0)
versions less than 1.5.0 get flagged as having a [vulnerability](https://www.mend.io/vulnerability-database/CVE-2024-5206?utm_source=JetBrains)

Also adds enhancement to catch the InconsistentVersionWarning raised by sklearn when reading a pickle file produced by a different version of sklearn and forces the classifier to be retrained.


gbeane:
rolling this into some other changes to pyproject.toml to support Python 3.13 

replaced by #81

PR #81, opened 2025-04-13
update pyproject.toml to support Python 3.13

gbeane:
pyproject.toml now allows installing with python >=3.10,<3.14

also updates:

* scikit-learn dependency ("1.3.0" -> ">=1.5.0,<2.0.0")
* pandas ("2.1.1" -> ">=2.2.2,<3.0")

also adds a check to force retraining if trying to load a cached classifier pickle file that was produced by a different version of sklearn


keithshep (approved):
looks good to me

PR #82, opened 2025-04-15
Refactor project.py

gbeane:
major refactoring of project.py.

I'm going to leave this open for comments at least until the end of the week, but likely for two weeks. 

This adds several new classes that are used by Project. These contain quite a bit of code that was cut and pasted from project.py, so there still might be room for improvement.  The most bang for your buck will be spending most of your time reviewing these new files:

* SettingsManager
* ProjectPaths
* PredictionManager
* FeatureManager
* VideoManager

The end result of this is that project.py length is reduced by about 300 lines, and Project.__init__ is reduced from around 150 lines to less than 25 (including a multi-line docstring).

Includes at least some unit tests for all new files (run with `pytest tests/project`) .

new files are formatted with `ruff format`
project.py reformatted with `ruff format`

There is more work to be done in project.py, but this is a step in the right direction.


tests/project/test_project_util.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+from src.jabs.project.project_utils import to_safe_name
+
+
+def test_to_safe_name():
+    """Test that to_safe_name converts strings to safe file names."""
+    assert to_safe_name("My Project") == "My_Project"
+    assert to_safe_name("Project@123") == "Project_123"
+    assert to_safe_name("File/Name") == "File_Name"
+    assert to_safe_name("  Leading and Trailing  ") == "Leading_and_Trailing"
```
SkepticRaven, line 12: This reminded me that somehow we had a project floating around with a leading ZWNBSP (U+FEFF) characters in the video filenames. Everything appeared to work fine, but I seem to recall those filenames looked strange in project.json.

SkepticRaven (approved):
Looks like a reasonable movement of functions out of project into better named sub-components.

src/jabs/project/settings_manager.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Save project properties & settings to the project file.
+        :param data: Dictionary with state information to save.
+        """
+        # Merge data with current metadata
+        self._project_info.update(data)
+        self._project_info["version"] = version_str()
+
+        # Save combined info to file
+        with self._paths.project_file.open(mode="w", newline="\n") as f:
```
bergsalex, line 51: You could consider IO error handling. 🤷

src/jabs/project/video_manager.py
```diff
@@ -0,0 +1,127 @@
...
+        self._total_project_identities = 0
+
+        self._initialize_videos(enable_video_check)
+
+    def _initialize_videos(self, enable_video_check):
+        """Initialize video-related data and perform checks."""
+        self._videos = self.get_videos(self._paths.project_dir)
+        self._videos.sort()
+
+        # self._validate_pose_files()
```
bergsalex, line 37: Commented code? Explain or remove?

src/jabs/project/video_manager.py
```diff
@@ -0,0 +1,127 @@
...
+
+from jabs.pose_estimation import get_pose_path, get_frames_from_file, open_pose_file
+from jabs.video_reader import VideoReader
+from jabs.video_reader.utilities import get_frame_count
+from .project_paths import ProjectPaths
+from .settings_manager import SettingsManager
+from .video_labels import VideoLabels
+
+
+class VideoManager:
```
bergsalex, line 13: Class level docstring?

src/jabs/project/video_manager.py
```diff
@@ -0,0 +1,127 @@
...
+        path = self._paths.annotations_dir / Path(video_filename).with_suffix(".json")
+
+        # if annotations already exist for this video file in the project open
+        # it, otherwise create a new empty VideoLabels
+        if path.exists():
+            with path.open() as f:
+                return VideoLabels.load(json.load(f))
+        else:
+            video_path = self._paths.project_dir / video_filename
+            nframes = get_frame_count(str(video_path))
```
bergsalex, line 71: This function potentially raises an `IOError`, does it make sense to do any handling of that here or is okay to just implicitly raise it?

src/jabs/project/video_manager.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _load_video_metadata(self):
+        """Load metadata for each video and calculate total identities."""
+        video_metadata = self._settings_manager.project_settings.get("video_files", {})
+        for video in self._videos:
+            vinfo = video_metadata.get(video, {})
+            nidentities = vinfo.get("identities")
+
+            if nidentities is None:
+                pose_file = open_pose_file(
```
bergsalex, line 99: Similar to my earlier comment. This function can raise a ValueError.

src/jabs/project/project_paths.py
```diff
@@ -0,0 +1,99 @@
...
+
+class ProjectPaths:
+    """
+    Class to manage project paths.
+    """
+
+    __JABS_DIR = "jabs"
+    __PROJECT_FILE = "project.json"
+
+    def __init__(self, base_path: Path, use_cache: bool = True):
```
bergsalex, line 12: Docstring? 😄

bergsalex (approved):

src/jabs/project/settings_manager.py
```diff
@@ -0,0 +1,100 @@
...
+        :param behavior: Behavior key to read.
+        :return: Dictionary of behavior metadata.
+        """
+        return self._project_info.get("behavior", {}).get(behavior, {})
+
+    def remove_behavior(self, behavior: str) -> None:
+        # remove from project settings
+        try:
+            del self._project_info["behavior"][behavior]
+            self.save_project_file({})
```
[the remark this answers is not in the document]
gbeane: yeah, I agree it is unintuitive.

save_project_file() takes a dict which is additional data to save in the project file (for example, if the user creates a new behavior in the project).  In this case we removed something from the project file and want to overwrite the copy on disk so I'm calling save_project_file() with no additional data. 

I'll change this to make the parameter optional (default to None) so this call would change to self.save_project_file()

src/jabs/project/video_manager.py
gbeane: good catch -- I think this is the result of some incomplete refactoring

keithshep (approved):

src/jabs/project/project_utils.py
```diff
@@ -0,0 +1,22 @@
+import re
+
+import h5py
+
+
+def to_safe_name(behavior: str) -> str:
+    """
+    Create a version of the given behavior name that
+    should be safe to use in filenames.
+    :param behavior: string behavior name
```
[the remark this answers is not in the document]
gbeane: at some point I'll go through and start updating docstrings to the google style (and also run the ruff formatter on the code), but not in this pull request. wonder if there is an automated tool for the docstring reformatting.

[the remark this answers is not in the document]
gbeane: added returns and raises.

src/jabs/project/video_manager.py
gbeane: it's okay to raise it. it should be caught and the GUI will display an alert message indicating why opening the project failed

in the future maybe we want to catch, do some logging, and then reraise

PR #85, opened 2025-05-06
add three keypoint lixit

gbeane:
adds support for lixit with three keypoints

in pose file, new style lixit has shape #lixit x 3 x 2 (old style is #lixit x 2)

keypoints are stored as tip, left, right

distance to tip is used for determining closest lixit, tip is still used for the distance from keypoint to lixit features. 

Adds additional feature (implemented in MouseLixitAngle), which is enabled in a project if all pose files have 3 keypoint lixit in the static objects. This is based on a feature Tuan was using in his classifier (along with distance from keypoints to lixit tip and window features).

Standard window features are computed on this new per-frame feature. 

also, for the files I touched updated type hints to remove old style typing.Dict, typing.Set, etc
ran ruff format on some of the touched files


src/jabs/feature_extraction/feature_group_base_class.py
```diff
@@ -8,7 +7,7 @@
 class FeatureGroup(abc.ABC):
 
     # to be defined in subclass
-    _features: typing.Dict[str, typing.Type[Feature]] = {}
+    _features: dict[str, Feature] = {}
```
keithshep, line 10: I didn't know you could parameterize the type if using `dict` over `typing.Dict`. Cool

src/jabs/feature_extraction/landmark_features/lixit.py
```diff
@@ -141,33 +173,78 @@ def per_frame(self, identity: int) -> typing.Dict:
...
+
+    @classmethod
+    def is_supported(cls, pose_version: int, static_objects: set[str], **kwargs) -> bool:
+        if super().is_supported(pose_version, static_objects, **kwargs):
+            # check additional attributes
+            if kwargs.get("lixit_keypoints") == 3:
+                return True
+        return False
+
+    def per_frame(self, identity: int) -> dict[str, np.ndarray]:
```
keithshep, line 236: I think this function would benefit from a pretty extensive comment hopefully including intuition/motivation for the calculations used

src/jabs/pose_estimation/__init__.py
```diff
@@ -88,3 +88,18 @@ def get_static_objects_in_file(path: Path):
...
+    """
+    points_per_lixit = 0
+    if get_pose_file_major_version(path) >= 5:
+        with h5py.File(path, 'r') as pose_h5:
+            if 'static_objects' in pose_h5 and 'lixit' in pose_h5['static_objects']:
+                if pose_h5['static_objects']['lixit'].ndim == 3:
+                    points_per_lixit = 3
+                else:
+                    points_per_lixit = 1
+    return points_per_lixit
```
keithshep, line 105: no newline at EOF :(

src/jabs/scripts/convert_parquet.py
```diff
@@ -137,8 +137,11 @@ def convert_data_frame(
         None
     """
 
-    identities = df["animal_id"].unique().tolist()
+    # we saw one parquet file with a single row with animal_id == 0
```
keithshep, line 140: yeah I think I've seen several animal_id == 0 with nothing there

keithshep (approved):

src/jabs/pose_estimation/__init__.py
gbeane: :(

src/jabs/feature_extraction/landmark_features/lixit.py
SkepticRaven: Agreed. I feel like this feature largely duplicates the information already present in lixit_bearings above. I would like to get more info on why this feature is being added.
Of particular interest is the implication that classifiers built with a 3-keypoint lixit will be incompatible with a 1-keypoint lixit (but not the other way around) exclusively for this one feature. Would this be an intended implication?

src/jabs/feature_extraction/features.py
SkepticRaven, line 1: Feature version not bumped

SkepticRaven (commented):
We may want to have a discussion about implications of changes to existing static object structure/associated features. Flexibility in adding in new things prohibits using newer classifiers on older data.

src/jabs/feature_extraction/features.py
gbeane: it would be good to add an easy way to disable this feature so that you can build a classifier that works for pose files with a single lixit keypoint using files with three lixit keypoints...

SkepticRaven: The other menu items for enabling/disabling features are located in the main_window.py ui as toggles.

gbeane:
@SkepticRaven @keithshep 
updated to use our centroid for computing the new lixit angle features
updated convert_parquet.py to drop their kpt_6

SkepticRaven (approved):

PR #86, opened 2025-05-09
add support for external ids 

gbeane:
uses optional dataset in pose file "posest/external_identity_mapping", which is a 1D uint32 array

Note: I added code to check for this dataset in all pose file versions >= 4.  Since this is completely optional, I didn't want to create yet another pose file version. 

This ID is displayed in the JABS GUI for both mouse labels in the video and the identity drop down selector:
image

Additionally, this is saved in the prediction hdf5 files as "predictions/external_identity_mapping" as well as in the json files JABS uses to store labels internally.  I did not bump the prediction version, but am willing to do so.

### DIFF
--- a/src/jabs/pose_estimation/pose_est_v4.py
+++ b/src/jabs/pose_estimation/pose_est_v4.py
@@ -21,7 +21,7 @@ class PoseEstimationV4(PoseEstimation):
     from this PoseEstimationV4 class because the only add additional optional information.
     """
 
-    __CACHE_FILE_VERSION = 3
+    __CACHE_FILE_VERSION = 4
 
     def __init__(self, file_path: Path,
                  cache_dir: typing.Optional[Path] = None,
@@ -70,6 +70,8 @@ class PoseEstimationV4(PoseEstimation):
                 all_confidence = pose_grp['confidence'][:]
                 id_mask = pose_grp['id_mask'][:]
                 instance_embed_id = pose_grp['instance_embed_id'][:]
+                if "external_identity_mapping" in pose_grp:
+                    self._external_identities = pose_grp["external_identity_mapping"][:].astype(int).tolist()
 
             self._num_frames = len(all_points)
             self._num_identities = np.max(np.ma.array(instance_embed_id[...], mask=id_mask[...]))
@@ -217,6 +219,8 @@ class PoseEstimationV4(PoseEstimation):
             self._num_identities = int(cache_h5.attrs['num_identities'])
             self._num_frames = int(cache_h5.attrs['num_frames'])
             self._identities = [*range(self._num_identities)]
+            if "external_identity_mapping" in pose_grp:
+                self._external_identities = pose_grp["external_identity_mapping"][:].astype(int).tolist()
 
             # get pixel size
             self._cm_per_pixel = pose_grp.attrs.get('cm_per_pixel', None)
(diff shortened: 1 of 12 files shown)


src/jabs/pose_estimation/pose_est_v4.py
```diff
@@ -217,6 +219,8 @@ def _load_from_cache(self):
             self._num_identities = int(cache_h5.attrs['num_identities'])
             self._num_frames = int(cache_h5.attrs['num_frames'])
             self._identities = [*range(self._num_identities)]
+            if "external_identity_mapping" in pose_grp:
+                self._external_identities = pose_grp["external_identity_mapping"][:].astype(int).tolist()
```
keithshep, line 223: can you explain why the `astype`. I'm sure there's a reason but not sure what

src/jabs/ui/central_widget.py
```diff
@@ -627,15 +634,37 @@ def _pixmap_clicked(self, event):
         handle event where user clicked on the video -- if they click
         on one of the mice, make that one active
         """
+        clicked_identity = None
```
keithshep, line 637: may be nice to have a separate `_get_identity_at_or_near(x, y)` or `_get_identity_at(x, y, tolerance)` that `_pixmap_clicked` uses

keithshep (approved):

src/jabs/pose_estimation/pose_est_v4.py
SkepticRaven, line 223: Would we want to convert to np.uint32 to match other changes?

src/jabs/project/prediction_manager.py
```diff
@@ -41,6 +41,7 @@ def write_predictions(
         probabilities,
         poses,
         classifier,
+        external_identities: list[int] | None = None,
```
SkepticRaven, line 44: Maybe np.uint32/unsigned int ([ref](https://numpy.org/doc/stable/user/basics.types.html#relationship-between-numpy-data-types-and-c-data-types))

src/jabs/project/video_labels.py
```diff
@@ -13,10 +13,11 @@ class VideoLabels:
 
     TODO stop using str for identities in method parameters, switch to int
     """
-    def __init__(self, filename, num_frames):
+    def __init__(self, filename, num_frames, external_identities: list[int] | None = None):
```
SkepticRaven, line 16: continued typing thread

src/jabs/project/video_manager.py
```diff
@@ -51,11 +51,12 @@ def videos(self):
     def total_project_identities(self) -> int:
         return self._total_project_identities
 
-    def load_video_labels(self, video_name):
+    def load_video_labels(self, video_name, external_identities: list[int] | None = None):
```
SkepticRaven, line 54: continued typing thread

SkepticRaven (approved):
Just a couple of related questions about typing. Want to make sure there isn't some inadvertent casting, particularly with potential negative numbers -> uint32.

src/jabs/pose_estimation/pose_est_v4.py
gbeane: added comment

gbeane: this is converting from the uint32 that they are stored as in the hdf5 file to int because uint32 is not serializable by the Python JSON serializer and I'm writing these out to the .json files that JABS uses to store labels

since Python int is boundless casting from an uint32 to an int shouldn't be an overflow issue

src/jabs/project/prediction_manager.py
gbeane: see comment above.  These are basically only used for labels in the GUI and also written out in the json label files and the prediction hdf5 files (I do convert back to np.uint32 for writing to the hdf5 file)

I could keep them as uint32 and override the JSON serializer to convert them to ints during serialization, but that seemed more trouble than just converting them when reading in from the pose file.

src/jabs/project/video_labels.py
gbeane: same response

src/jabs/project/video_manager.py
gbeane: same response

PR #88, opened 2025-05-11
Update numpy

gbeane:
updates to numpy ^2.0.0 (>=2.0.0 < 3.0.0)
*also requires pandas upgrade to ^2.2.2 due to change in numpy C API resulting in a binary incompatibility with older versions of pandas

The primary reasoning behind this change is that binary distributions of older versions of numpy are not available for Python 3.13 on most platforms, so installation required compiling numpy from source (which requires a compiler, which some users will not have).

Numpy version 2 is well supported for Python 3.10-3.13 and binary distributions are available for Windows, Mac, and Linux


keithshep (approved):
looks good

PR #97, opened 2025-05-13
persist acceptance of license agreement

### LINKED ISSUE #92: remember license agreement
It would be nice to have JABS remember license agreement so that it only happens once.

gbeane:
closes #92 

Uses QSettings to persist the user's acceptance of the license agreement 
also saves the JABS version at the time of license acceptance so we can force re-acceptance in the future if we want (for example, if license changes in version X we could force anyone that accepted license terms prior to version X to re-confirm)

Also use QSettings to keep track of the most recent projects.  Adds a File->Open Recent menu that is populated with the most recently opened projects (up to 5)

image


keithshep (approved):
all looks good to me

PR #100, opened 2025-05-14
improve performance when training projects with lots of unlabeled videos

gbeane:
When training, jabs was iterating over each video in the project and even if there were no labels it would open the pose file and open the features for each identity in the video

This change skips entire videos without labels entirely (without opening pose or features). For videos where some identities have labels and some don't, the pose file is read in but it will skip over reading the features hdf5 files for the identities without labels

also fix some tests, skip some tests that have been broken for a long time


src/jabs/project/project.py
```diff
@@ -413,6 +432,17 @@ def get_labeled_features(self, behavior=None, progress_callable=None):
             for identity in pose_est.identities:
                 group_mapping[group_id] = {"video": video, "identity": identity}
 
+                labels = video_labels.get_track_labels(str(identity), behavior).get_labels()
+
+                # if there are no labels for this identity, skip it
+                if (
```
keithshep, line 438: feel free to ignore this. I think you can simplify as:

```
# if there are no labels for this identity, skip it
if np.count_nonzero(
    (labels == TrackLabels.Label.BEHAVIOR) | (labels == TrackLabels.Label.NOT_BEHAVIOR)
) == 0:
    if progress_callable is not None:
        progress_callable()
    continue
```

keithshep (approved):

SkepticRaven (approved):

PR #107, opened 2025-05-16
add hdf5 compression

gbeane:
adds compression to hdf5 files:
* features hdf5 files
* cached v4+ transformed pose hdf5 files

also fixes some type hints and removed an numpy array that was created but never used

### DIFF
--- a/src/jabs/feature_extraction/features.py
+++ b/src/jabs/feature_extraction/features.py
@@ -5,6 +5,7 @@ import numpy as np
 import pandas as pd
 
 import jabs.project.track_labels
+from jabs.constants import COMPRESSION, COMPRESSION_OPTS_DEFAULT
 from jabs.pose_estimation import PoseEstimation, PoseHashException
 
 # import feature modules
@@ -39,9 +40,6 @@ _WINDOW_FILTERS = {
     'fft': list(Feature._signal_operations.keys()),
 }
 
-_COMPRESSION = "gzip"
-_COMPRESSION_OPTS_DEFAULT = 6
-
 
 class FeatureVersionException(Exception):
     pass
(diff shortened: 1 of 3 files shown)


src/jabs/feature_extraction/features.py
```diff
@@ -242,37 +242,38 @@ def __save_per_frame(self):
             if self._distance_scale_factor is not None:
                 features_h5.attrs['distance_scale_factor'] = self._distance_scale_factor
             features_h5.attrs['pose_hash'] = self._pose_hash
-            features_h5.create_dataset('frame_valid', data=self._frame_valid)
+            features_h5.create_dataset('frame_valid', data=self._frame_valid, compression="gzip")
```
SkepticRaven, line 245: For the segmentation data, I also use `compression_opts=9`, which could also be used here if we needed to squeeze even more space.

SkepticRaven (approved):
It might be useful to see what sort of tradeoff we get for compression vs read/write speeds. If its a big enough hit on reads, we could consider exposing an option for `generate_features.py` and other CLI interfaces.

src/jabs/feature_extraction/features.py
gbeane: yeah, I went with the default (which I think is 4) but maybe I'll add in the compression_opts parameter so it would be easier to change this in the future

gbeane: @SkepticRaven updated PR to parameterize this. Made the default 6, which can be overridden.

src/jabs/pose_estimation/pose_est_v4.py
keithshep, line 1: any reason why `compression_opts` are used elsewhere but not here?

keithshep (approved):

PR #110, opened 2025-05-16
use docconvert to change re-Structured Text docstring format to google

gbeane:
the Kumar Lab is standardizing on google format for Python docstrings 

most of the existing code base used re-Structured Text (for formatted docstrings, many docstrings are just an unstructured string comment)

This pull request uses the docconvert package to automatically reformat docstrings. 

The auto formatter might have gotten a few things wrong for docstrings that were almost but not quite re-Structured Text, but we'll correct those as we find them.

### DIFF
--- a/src/jabs/classifier/classifier.py
+++ b/src/jabs/classifier/classifier.py
@@ -56,10 +56,9 @@ class Classifier:
 
     def __init__(self, classifier=ClassifierType.RANDOM_FOREST, n_jobs=1):
         """
-        :param classifier: type of classifier to use. Must be ClassifierType
-        :param n_jobs: number of jobs to use for classifiers that support
-        this parameter for parallelism
-        enum value. Defaults to ClassifierType.RANDOM_FOREST
+        Args:
+            classifier: type of classifier to use. Must be ClassifierType enum value. Defaults to ClassifierType.RANDOM_FOREST
+            n_jobs: number of jobs to use for classifiers that support this parameter for parallelism
         """
 
         self._classifier_type = classifier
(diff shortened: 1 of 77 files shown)


src/jabs/classifier/classifier.py
```diff
@@ -56,10 +56,9 @@ class Classifier:
 
     def __init__(self, classifier=ClassifierType.RANDOM_FOREST, n_jobs=1):
         """
-        :param classifier: type of classifier to use. Must be ClassifierType
-        :param n_jobs: number of jobs to use for classifiers that support
-        this parameter for parallelism
-        enum value. Defaults to ClassifierType.RANDOM_FOREST
+        Args:
+            classifier: type of classifier to use. Must be ClassifierType enum value. Defaults to ClassifierType.RANDOM_FOREST
+            n_jobs: number of jobs to use for classifiers that support this parameter for parallelism
```
keithshep, line 61: I'm guessing this is something that you manually fixed? Otherwise I'm confused how this happened

src/jabs/feature_extraction/base_features/angles.py
```diff
@@ -50,13 +49,15 @@ def window(self, identity: int, window_size: int,
...
             a: np.ndarray, b: np.ndarray, c: np.ndarray
     ) -> np.ndarray:
-        """
-        compute angles for a set of points
-        :param a: array of point coordinates
-        :param b: array of vertex point coordinates
-        :param c: array of point coordinates
-        :return: array containing angles, in degrees, formed from the lines
-        ab and ba for each row in a, b, and c with range [0, 360)
+        """compute angles for a set of points
```
keithshep, line 52: I think it reads nicer with the newline after `"""`. Is there a formatter option to keep that newline?

keithshep (approved):

src/jabs/classifier/classifier.py
gbeane: yes sir

SkepticRaven: Likely parameter being added and docstring edited in the middle of classifier arg rather than after.

PR #120, opened 2025-05-20
fix ruff check errors

gbeane:
first round of ruff linter error fixes

src/jabs/utils
src/jabs/version
src/jabs/video_reader
src/jabs/types

also sets target version to py310 since we want to ensure code is compatible with Python 3.10 or greater. Older versions are no longer supported.


keithshep (approved):
all looks good to me

PR #121, opened 2025-05-21
fix ruff check linting errors in src/jabs/ui

gbeane:
round 2 of fixing `ruff check` linting errors:

addresses ruff linting errors in /src/jabs/ui

### DIFF
--- a/src/jabs/ui/central_widget.py
+++ b/src/jabs/ui/central_widget.py
@@ -1,7 +1,7 @@
 import sys
 
 import numpy as np
-from PySide6 import QtWidgets, QtCore, QtGui
+from PySide6 import QtCore, QtGui, QtWidgets
 from shapely.geometry import Point
 
 import jabs.feature_extraction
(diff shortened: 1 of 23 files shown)


src/jabs/ui/central_widget.py
```diff
@@ -105,17 +105,20 @@ def __init__(self, *args, **kwargs):
...
-            ):
+            if event.type() == QtCore.QEvent.Type.Close:
                 event.accept()
                 return True
+            elif event.type() == QtCore.QEvent.Type.KeyPress and isinstance(
+                event, QtGui.QKeyEvent
+            ):
+                if event.key() == QtCore.Qt.Key.Key_Escape:
+                    event.accept()
+                    return True
```
keithshep, line 121: This seems gnarly to me. Can we have a single elif rather than elif-if

src/jabs/ui/main_control_widget.py
```diff
@@ -1,16 +1,18 @@
 """Project-level controls for classifiers.
 
-TODO:
```
keithshep, line 3: this seems to me like an unhelpful formatting rule is forcing this

keithshep (approved):
Just one real comment. Up to you what you want to do with it

src/jabs/ui/central_widget.py
gbeane: how about this

```
    def eventFilter(self, source, event):
        """filter events emitted by progress dialog"""
        if source != self._progress_dialog:
            pass
        else:
            if event.type() == QtCore.QEvent.Type.Close or (event.type() == QtCore.QEvent.Type.KeyPress and isinstance(
                event, QtGui.QKeyEvent) and event.key() == QtCore.Qt.Key.Key_Escape):
                event.accept()
                return True
        return super().eventFilter(source, event)

```

PR #122, opened 2025-05-22
fix ruff errors in src/jabs/scripts

gbeane:
fixes ruff check errors for files in src/jabs/scripts directory

### DIFF
--- a/src/jabs/scripts/convert_parquet.py
+++ b/src/jabs/scripts/convert_parquet.py
@@ -79,8 +79,8 @@ import sys
 from pathlib import Path
 
 import h5py
-import pandas as pd
 import numpy as np
+import pandas as pd
 from argparse_formatter import FlexiFormatter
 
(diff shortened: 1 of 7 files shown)


src/jabs/scripts/convert_parquet.py
```diff
@@ -221,7 +225,7 @@ def read_lixit_csv(path: Path) -> dict[str, tuple[float, float]]:
     right_side_x = []
     right_side_y = []
 
-    with open(path, "r") as f:
```
gbeane, line 224: I'm not sure I agree with enforcing this:  https://docs.astral.sh/ruff/rules/redundant-open-modes/

ruff wants to remove "r"  because it's the default and considered redundant

keithshep: I would vote against that rule. I think it's nice to have the "r" from a code-is-documentation perspective

keithshep (approved):

PR #123, opened 2025-05-22
fix ruff check errors in src/jabs/project

gbeane:
fixes ruff check errors

also fixes some docstrings that were either formatted poorly by the automated doctoring style formatting or were lacking in content


keithshep (approved):
looks good to me

PR #125, opened 2025-05-22
Fix project import and label serialization

gbeane:


src/jabs/project/track_labels.py
```diff
@@ -234,7 +234,9 @@ def _array_to_blocks(cls, array: np.ndarray) -> list[dict]:
                     {
                         "start": block_start,
                         "end": block_start + count - 1,
-                        "present": val == cls.Label.BEHAVIOR,
+                        # note: val == cls.Label.BEHAVIOR returns a numpy bool, which is not json serializable
+                        # the True if ... else False forces it to be a python bool
+                        "present": True if val == cls.Label.BEHAVIOR else False,  # noqa: SIM210
```
keithshep, line 239: I think `bool(val == cls.Label.BEHAVIOR)` might be better here (unless it doesn't do it for some reason)

keithshep (approved):

src/jabs/project/track_labels.py
gbeane: yeah, that would probably be a better option

PR #127, opened 2025-05-28
src/jabs now passes all ruff linting checks

gbeane:
last batch of changes so that all files in src/jabs pass `ruff check` 

lots of the changes involve adding docstrings or annotating class variables with typing.ClassVar

### DIFF
--- a/src/jabs/feature_extraction/base_features/angles.py
+++ b/src/jabs/feature_extraction/base_features/angles.py
@@ -23,7 +23,7 @@ class Angles(Feature):
         super().__init__(poses, pixel_scale)
         self._num_angles = len(AngleIndex)
 
-    def per_frame(self, identity: int) -> np.ndarray:
+    def per_frame(self, identity: int) -> dict[str, np.ndarray]:
         """compute the value of the per frame features for a specific identity"""
         values = {}
 
--- a/src/jabs/feature_extraction/base_features/angles.py
+++ b/src/jabs/feature_extraction/base_features/angles.py
@@ -14,7 +14,7 @@ class Angles(Feature):
     _name = "angles"
 
     # override for circular values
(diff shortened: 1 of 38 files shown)


src/jabs/feature_extraction/base_features/angles.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, poses: PoseEstimation, pixel_scale: float):
         self._num_angles = len(AngleIndex)
 
     def per_frame(self, identity: int) -> np.ndarray:
```
bergsalex, line 26: This type seems to be wrong. I think it's a `dict[str, np.ndarray]`.

bergsalex (approved):

src/jabs/feature_extraction/feature_base_class.py
```diff
@@ -20,10 +19,10 @@ class Feature(abc.ABC):
 
     # requirements for this feature to be available
     _min_pose = 2
-    _static_objects = []
+    _static_objects: typing.ClassVar[list[str]] = []
 
     # NOTE: Circular values need to override this as well as the window()
-    _window_operations = {
+    _window_operations: typing.ClassVar[dict[str, typing.Callable]] = {
```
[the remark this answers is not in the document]
gbeane: _window_operations is a dictionary that maps a feature name to a function that is used to compute it, so typing.Callable should be right.  

It looks like it is a mistake where it's annotated as `typing.ClassVar[dict[str, np.ndarray]]` and I will fix.

src/jabs/feature_extraction/base_features/angles.py
gbeane: yup, there are a bunch of these that were wrong that I fixed (these functions used to return ndarrays but then changed to dicts and most of the return type hints were not properly updated)

PR #135, opened 2025-06-04
disable select mode if label timeline is hidden

gbeane:
if the user hides the label timeline (selecting View->Timeline->Predictions instead of View->Timeline->Labels & Predictions or View->Timeline->Labels) then they can't see what they are selecting and labeling. 

This disables selecting and labeling frames if the label timeline is hidden

Also prevents using the z, x, and c buttons to enter select mode if the select frames button is disabled.

### DIFF
--- a/src/jabs/ui/central_widget.py
+++ b/src/jabs/ui/central_widget.py
@@ -842,15 +842,14 @@ class CentralWidget(QtWidgets.QWidget):
             self._controls.select_button_set_checked(False)
             self._stacked_timeline.clear_selection()
 
-        # if no video is loaded, or if there are no identities, disable the select button
-        if self._loaded_video is None or (
-            self._pose_est is not None and self._pose_est.num_identities == 0
+        # disable select frames button if no video is loaded, there are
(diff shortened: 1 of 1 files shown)


src/jabs/ui/central_widget.py
```diff
@@ -833,3 +826,31 @@ def _decrement_identity_index(self):
...
+            self._pose_est is not None and self._pose_est.num_identities == 0
+        ):
+            disable_select_button()
+
+        if self._stacked_timeline.view_mode == self._stacked_timeline.view_mode.PREDICTIONS:
+            # If the view mode is predictions, disable the select button because there is
+            # no way for the user to see what they are labeling
+            disable_select_button()
+        else:
+            self._controls.select_button_enabled = True
```
keithshep, line 856: recommend `elif` for clarity. Alternatively remove nested function and have a single `if` with longer condition chain.

keithshep (approved):

src/jabs/ui/central_widget.py
gbeane: agree

PR #137, opened 2025-06-06
fix exception loading pose file with no identities

gbeane:
I had a project that included a few videos where the mice were hidden under nesting material for the entire 1 minute segment.  This causes an exception and prevented JABS from loading the project. 

This change makes it so that JABS won't raise an exception when reading the pose file so the project can still be opened. 

Not sure if there might be other bugs when training / predicting on this project, but for now this will at least allow us to open the project and label other videos.


keithshep (approved):
looks good

PR #141, opened 2025-06-10
Allow labeling frames with dropped identities

gbeane:
Internal JIRA: NVSN-26

Allows saving labels over frames with dropped identities. There is a new key in the JSON file that the labels are saved in called "unfragmented_labels", which contains intervals that might include frames with no identity. The original "labels" key still contains intervals for frames with identity only. 

In the GUI, labels over frames with dropped identity are shown using a semi-transparent color in the Label Timeline so the identity gap fill pattern is still visible:

identity_gaps_with_label

example json file with new key:
```
{
  "file": "cage_2453.2025-02-12.00.10.mp4",
  "num_frames": 1800,
  "labels": {
    "0": {
      "Drinking": [
        {
          "start": 6,
          "end": 53,
          "present": false
        },
        {
          "start": 55,
          "end": 57,
          "present": false
        },
        {
          "start": 59,
          "end": 59,
          "present": false
        },
        {
          "start": 88,
          "end": 88,
          "present": false
        },
        {
          "start": 97,
          "end": 97,
          "present": false
        },
        {
          "start": 100,
          "end": 100,
          "present": false
        },
        {
          "start": 102,
          "end": 102,
          "present": false
        },
        {
          "start": 104,
          "end": 160,
          "present": false
        }
      ]
    },
    "1": {},
    "2": {}
  },
  "unfragmented_labels": {
    "0": {
      "Drinking": [
        {
          "start": 0,
          "end": 62,
          "present": false
        },
        {
          "start": 81,
          "end": 160,
          "present": false
        }
      ]
    },
    "1": {},
    "2": {}
  },
  "labeler": "gbeane"
}
```


keithshep (approved):
nice one. I tried it out with my project and ran unit tests

PR #143, opened 2025-06-10
add select all feature

keithshep:
ctrl+a/cmd+a will now select all frames in the current timeline for labeling

### DIFF
--- a/src/jabs/ui/main_window.py
+++ b/src/jabs/ui/main_window.py
@@ -274,6 +274,12 @@ class MainWindow(QtWidgets.QMainWindow):
         self.enable_segmentation_features.triggered.connect(self._toggle_segmentation_features)
         feature_menu.addAction(self.enable_segmentation_features)
 
+        # select all action
+        select_all_action = QtGui.QAction(self)
+        select_all_action.setShortcut(QtGui.QKeySequence.SelectAll)
+        select_all_action.triggered.connect(self._handle_select_all)
+        self.addAction(select_all_action)
+
         # playlist widget added to dock on left side of main window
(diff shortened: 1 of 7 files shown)


src/jabs/ui/main_window.py
```diff
@@ -275,6 +275,12 @@ def __init__(self, app_name: str, app_name_long: str, *args, **kwargs):
         self.enable_segmentation_features.triggered.connect(self._toggle_segmentation_features)
         feature_menu.addAction(self.enable_segmentation_features)
 
+        # select all action
+        select_all_action = QtGui.QAction(self)
+        select_all_action.setShortcut(QtGui.QKeySequence.SelectAll)
```
gbeane, line 280: I think in PySide6 this should be `QtGui.QKeySequence.StandardKey.SelectAll`

`QtGui.QKeySequence.SelectAll` works, I think for some compatibility with previous versions of Pyside, but gives an IDE warning: `Unresolved attribute reference 'SelectAll' for class 'QKeySequence'`. 

same with QtQui.QKeySequence.Find elsewhere in this file.  I had a bunch of instances of things like that from when we were using PySide2 and PyQt and i've been slowly trying to get everything updated to the preferred package paths for PySide6

https://doc.qt.io/qtforpython-6/PySide6/QtGui/QKeySequence.html#PySide6.QtGui.QKeySequence.StandardKey

gbeane (approved):
approved, but consider updating to use QtGui.KeySequence.StandardKey.SelectAll

src/jabs/ui/main_window.py
keithshep: cool. fixed it

PR #144, opened 2025-06-11
make the GUI handle exceptions raised in training and classification threads

### LINKED ISSUE #72: Better error handling in training and prediction threads
Right now if an exception is thrown in either then training or prediction threads, the execution of the thread stops but the modal progress bar stays open, preventing the user from interacting with the application requiring a force quit. 

We need better error handling to catch exceptions, display appropriate error messages to the user, and cleanup thread and progress bar

gbeane:
previously, if an exception were raised during training or classification, the GUI would get stuck with the progress dialog open. The only remedy was to force quit the application. This was also confusing to users, because unless they looked in the terminal and saw the exception stack trace, they had no indication something was wrong other than the progress bar not moving for a long period of time. 

This change dismisses the progress bar and displays an alert dialog with a message about the exception that occurred. Also the stack trace is printed to the terminal to aid with debugging. In the future we could have CentralWidget treat different types of exceptions differently -- for example if there is a problem with data we could give the user a different error message (and not print the stack trace) vs if an unhandled exception is raised due to a bug in the code (need stack trace for debugging). 

Fixes #72

### DIFF
--- a/src/jabs/ui/central_widget.py
+++ b/src/jabs/ui/central_widget.py
@@ -614,7 +614,7 @@ class CentralWidget(QtWidgets.QWidget):
     def _classify_thread_error_callback(self, error: Exception) -> None:
         """handle an error in the classification thread"""
         self._print_exception(error)
-        self._cleanup_training_thread()
+        self._cleanup_classify_thread()
         self._cleanup_progress_dialog()
         self.status_message.emit("Classification Failed", 3000)
         self._controls.train_button_enabled = True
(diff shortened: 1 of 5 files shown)


src/jabs/ui/central_widget.py
```diff
@@ -590,11 +595,57 @@ def _train_button_clicked(self) -> None:
...
+        self.status_message.emit("Training Failed", 3000)
+        self._controls.classify_button_enabled = False
+        QtWidgets.QMessageBox.critical(
+            self, "Error", f"An exception occurred during training:\n{error}"
+        )
+
+    def _classify_thread_error_callback(self, error: Exception) -> None:
+        """handle an error in the classification thread"""
+        self._print_exception(error)
+        self._cleanup_training_thread()
```
gbeane, line 617: oops. this is wrong -- need to create a _cleanup_classify_thread method to call here

SkepticRaven (approved):

PR #148, opened 2025-06-18
implement method in PlayerWidget to play back a range of frames

gbeane:
implement method in PlayerWidget to play back a range of frames

uses new functionality to implement "play current bout" keyboard controls.

shift-space: plays current labeled bout (only if there is a label for the current frame)
alt-space: plays current predicted bout  (only if there is a prediction at the current frame)

The main reason behind implementing this is to use the play_range for future functionality to play through search hits.

### DIFF
--- a/src/jabs/ui/player_widget/player_widget.py
+++ b/src/jabs/ui/player_widget/player_widget.py
@@ -1,4 +1,5 @@
 import enum
+from dataclasses import dataclass
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


src/jabs/ui/player_widget/player_widget.py
```diff
@@ -13,14 +14,28 @@
 _SPEED_VALUES = [0.5, 1, 2, 4]
 
 
+@dataclass(frozen=True)
+class PlaybackRange:
+    """Dataclass to represent a playback range in the video."""
+
+    start: int
+    end: int
```
keithshep, line 22: probably worth mentioning frame units in pydoc or in member var name

keithshep (approved):
looks good

PR #152, opened 2025-06-24
adds field for filtering the list of videos using a partial string match

gbeane:
lets user filter the list of videos in a project by a simple string partial match

e.g. if a user types "foo" in the search box, all video names that contain the substring "foo" will be shown in the video list

This comparison is case insensitive 

filtering the video list does not change the selected video, even if the selected video is now hidden in the list.  The user needs to click on a new video in the filtered results to change the loaded video. 

Also, makes selected video more obvious

### DIFF
--- a/src/jabs/ui/video_list_widget.py
+++ b/src/jabs/ui/video_list_widget.py
@@ -111,7 +111,7 @@ class VideoListDockWidget(QtWidgets.QDockWidget):
         """Filter the video list based on the text entered in the filter box."""
         for i in range(self._file_list.count()):
             item = self._file_list.item(i)
-            item.setHidden(text.lower() not in item.text().lower())
+            item.setHidden(text.strip().lower() not in item.text().lower())
 
     def set_project(self, project):
         """Update the video list with the active project's videos and select first video in list."""


src/jabs/ui/video_list_widget.py
```diff
@@ -1,16 +1,59 @@
...
+                # Set pen to highlighted text color
+                painter.setPen(option.palette.color(QtGui.QPalette.ColorRole.HighlightedText))  # type: ignore[attr-defined]
+
+                # Draw the text manually
+                text = index.data(QtCore.Qt.ItemDataRole.DisplayRole)
+                rect = option.rect.adjusted(4, 0, -4, 0)  # type: ignore[attr-defined]
+                painter.drawText(rect, QtCore.Qt.AlignmentFlag.AlignVCenter, text)
+                painter.restore()
+            else:
+                super().paint(painter, option, index)
```
keithshep, line 49: I don't know if you maybe did already but I'm wondering if we should give this a try with win using dark mode/light mode before merge

src/jabs/ui/video_list_widget.py
```diff
@@ -44,21 +100,30 @@ def _selection_changed(self, current, _):
...
             video = current.data(QtCore.Qt.ItemDataRole.UserRole)
             self.selectionChanged.emit(video)
+        else:
+            self.selectionChanged.emit("")
+
+    def _filter_list(self, text):
+        """Filter the video list based on the text entered in the filter box."""
+        for i in range(self._file_list.count()):
+            item = self._file_list.item(i)
+            item.setHidden(text.lower() not in item.text().lower())
```
keithshep, line 110: wondering if we should strip text. Or if not then maybe `text = text if text.strip() else ''` to avoid user confusion in the case that there is whitespace left in search box? Not convinced on this suggestion though. Ignore if you don't like it.

keithshep (approved):

src/jabs/ui/video_list_widget.py
keithshep: On Windows though? Or maybe it doesn't matter if it's using the same UI rendering for both platforms? (wasn't sure about that part)

gbeane: I'll check on Windows when I'm in the office tomorrow.  We can hold off on merging until then.

PR #156, opened 2025-06-26
add keyboard controls for switching to previous or next video

gbeane:
uses comma and period key presses to switch to previous or next video


keithshep (approved):
looks good

PR #164, opened 2025-07-02
add project merge

### LINKED ISSUE #103: Project merge capability
It would be useful to have a capability to be able to merge labels from multiple projects. This could be project merge or label import.

gbeane:
closes #103


src/jabs/project/track_labels.py
```diff
@@ -238,3 +240,41 @@ def _array_to_blocks(cls, array: np.ndarray) -> list[dict]:
                 )
             block_start += count
         return blocks
+
+    def merge(self, other: "TrackLabels", strategy: MergeStrategy) -> None:
```
keithshep, line 244: Do we lose unfragmented labels with this approach? May need to take an interval based approach to merge if so.

gbeane: both the original (fragmented) and the new unfragmented intervals are generated from a numpy array

I merge the arrays, then serialize to json, so we should get correct fragmented and unfragmented intervals based on the new merged labels.

PR #172, opened 2025-07-16
deal with more errors reading cached convex hulls

gbeane:
handle more types of errors when trying to read cached convex hull arrays

also uses joblib for reading/writing pickle file (used elsewhere in JABS rather than the pickle module because it handles numpy arrays better. changing here just for consistency )

### DIFF
--- a/src/jabs/pose_estimation/pose_est.py
+++ b/src/jabs/pose_estimation/pose_est.py
@@ -4,6 +4,7 @@ from abc import ABC, abstractmethod
 from pathlib import Path
 
 import h5py
+import joblib
 import numpy as np
 from shapely.geometry import MultiPoint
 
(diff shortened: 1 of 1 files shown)


src/jabs/pose_estimation/pose_est.py
```diff
@@ -255,8 +256,8 @@ def get_identity_convex_hulls(self, identity):
 
                 try:
                     with path.open("rb") as f:
-                        convex_hulls = pickle.load(f)
-                except OSError:
+                        convex_hulls = joblib.load(f)
+                except (OSError, EOFError, pickle.UnpicklingError):
```
keithshep, line 260: I know it's generally bad practice but wondering if it makes sense to catch Exception here? (Not convinced it's right to do, really just wondering)

PR #173, opened 2025-07-17
fix issue 167

### LINKED ISSUE #167: Missing social features
Docs suggest 9 social distance features, however only 3 are appearing in the cache/classifiers.
Cursory inspection suggests it was introduced in the feature name propagation changes I made in `compute_pairwise_social_distances` [here](https://github.com/KumarLabJax/JABS-behavior-classifier/blob/main/src/jabs/feature_extraction/social_features/social_distance.py#L184). The loop should have probably been `len(values)` (9) rather than `len(social_points)` (3).

That being said, this function is currently coupled with `_compute_pairwise_social_distances` below it with assumed ordering of names and should probably be appropriately decoupled if possible.

gbeane:
fixes #167

### DIFF
--- a/src/jabs/feature_extraction/social_features/social_distance.py
+++ b/src/jabs/feature_extraction/social_features/social_distance.py
@@ -170,23 +170,3 @@ class ClosestIdentityInfo:
                     return_dict[key][frame] = math.dist(points[p1.value], closest_points[p2.value])
 
         return return_dict
-
-    @staticmethod
-    def _compute_social_pairwise_distance(points1, points2):
-        """compute distances between all pairs of points
-
-        Args:
-            points1: 1st collection of points
-            points2: 2nd collection of points
-
-        Returns:
-            list of distances between all pairwise combinations of
-            points from points1 and points2
-        """
-        distances = []
-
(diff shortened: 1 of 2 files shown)


src/jabs/feature_extraction/social_features/social_distance.py
```diff
@@ -165,34 +164,9 @@ def compute_pairwise_social_distances(
 
             closest_points, _ = self._poses.get_points(frame, closest_id, self._pixel_scale)
 
-            values[frame] = self._compute_social_pairwise_distance(
-                points[social_pt_indexes, ...], closest_points[social_pt_indexes, ...]
-            )
+            for p1 in social_points:
+                for p2 in social_points:
+                    key = f"social dist. {p1.name}-{p2.name}"
+                    return_dict[key][frame] = math.dist(points[p1.value], closest_points[p2.value])
```
keithshep, line 170: Not really for this PR but maybe something to consider for future performance improvement work: could we replace `math.dist` with `np.linalg.norm`

keithshep (approved):

SkepticRaven (approved):

PR #181, opened 2025-07-30
Update for singularity

gbeane:
some updates to get the JABS CLI singularity image to build cleanly

Note: image is still based on Python 3.10 but I think we should consider moving to Python 3.13 to take advantage of work that has been done to improve Python performance


SkepticRaven (approved):
Looks like there's a `python:3.13-bookworm` which should be trivial to swap to. No opposition to include it here, as long as the container can build!

Trixie is in testing right now, but it looks like at least the python docker starts are supporting the "current 2" debians.

PR #194, opened 2025-08-13
fix exception when enabling/disabling features in Feature menu

gbeane:
fixes exception when enabling/disabling features using the Feature menu

The bug was related to some refactoring that moved some functionality out of the Project class and into a new class called SettingsManager


keithshep (approved):
looks good

SkepticRaven (approved):

PR #202, opened 2025-08-18
Add contrast adjustment control

gbeane:
adds control to video controls overlay to allow user to adjust contrast using slider to choose a value from 1.0 (original contrast) to 2.0 (high contrast). 

image


SkepticRaven (approved):
Excellent!

PR #203, opened 2025-08-20
update xgboost version

gbeane:
we were pinned to a specific XGBoost version from June 2023.  This update allows all version 2.x.  This seems safe. It runs and produces similar CV metrics for some test classifiers. 

we might want to consider allowing 3.x, which came out March 2025


keithshep (approved):
looks good

bergsalex (approved):

PR #222, opened 2025-09-17
Rename behavior utility

gbeane:
Note: branch name is misleading, should be *rename-behavior-util*

add cli utility for renaming a behavior in an existing JABS project

usage:

`jabs-cli rename-behavior <project dir> <old behavior name> <new behavior name>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ After installation, the following commands are available in your environment:
 - `jabs-classify` — run a trained classifier  
 - `jabs-stats` — print accuracy statistics for a classifier  
 - `jabs-export-training` — export training data from an existing JABS project
+- `jabs-cli` - collection of smaller command line utilities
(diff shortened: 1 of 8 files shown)


gbeane:
@SkepticRaven 

> I like the new `is_valid_project_directory` method. I see a lot of project folders inside project folders because someone went into the jabs folder.

Yeah, I need to expand its use now -- especially for any command line tool that expects to operate on an existing JABS project directory so we avoid creating `jabs` directories if the user passes an incorrect path.   For the GUI, I want users to be able to open a directory that's never been opened in JABS before but we could prompt for confirmation if it's not already a JABS project directory (or maybe for the GUI it would be more useful to check that the selected directory actually contains video/pose pairs)

bergsalex (approved):

PR #227, opened 2025-09-25
Implement bounding boxes

gbeane:
add pose v8 support
implement bounding box overlay (preferred identity overlay)
  will default to bounding box overlay and then fall back to floating id overlay if a pose file is opened without bounding boxes

image

### DIFF
--- a/src/jabs/pose_estimation/pose_est_v6.py
+++ b/src/jabs/pose_estimation/pose_est_v6.py
@@ -47,14 +47,16 @@ class PoseEstimationV6(PoseEstimationV5):
             # transpose seg_data similar to the way the points are transposed.
 
         # sort the segmentation data
-        self._segmentation_dict["seg_data"] = self._segmentation_sort(
-            self._segmentation_dict["seg_data"],
-            self._segmentation_dict["longterm_seg_id"],
-        )
-        self._segmentation_dict["seg_external_flag"] = self._segmentation_sort(
-            self._segmentation_dict["seg_external_flag"],
-            self._segmentation_dict["longterm_seg_id"],
-        )
+        if self._segmentation_dict["seg_data"] is not None:
+            self._segmentation_dict["seg_data"] = self._segmentation_sort(
+                self._segmentation_dict["seg_data"],
+                self._segmentation_dict["longterm_seg_id"],
+            )
+        if self._segmentation_dict["seg_external_flag"] is not None:
+            self._segmentation_dict["seg_external_flag"] = self._segmentation_sort(
+                self._segmentation_dict["seg_external_flag"],
+                self._segmentation_dict["longterm_seg_id"],
+            )
 
     def get_seg_id(self, frame_index: int, identity: int) -> np.ndarray[Any, Any] | None:
         """get segmentation for a given frame and identity."""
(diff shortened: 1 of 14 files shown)


SkepticRaven (approved):

src/jabs/pose_estimation/pose_est_v6.py
```diff
@@ -47,14 +47,16 @@ def __init__(self, file_path: Path, cache_dir: Path | None = None, fps: int = 30
...
+        if self._segmentation_dict["seg_data"] is not None:
+            self._segmentation_dict["seg_data"] = self._segmentation_sort(
+                self._segmentation_dict["seg_data"],
+                self._segmentation_dict["longterm_seg_id"],
+            )
+        if self._segmentation_dict["seg_external_flag"] is not None:
+            self._segmentation_dict["seg_external_flag"] = self._segmentation_sort(
+                self._segmentation_dict["seg_external_flag"],
+                self._segmentation_dict["longterm_seg_id"],
+            )
```
keithshep, line 59: is this safe or should it be `if self._segmentation_dict.get("seg_data") is not None:`?

src/jabs/pose_estimation/pose_est_v8.py
```diff
@@ -0,0 +1,160 @@
...
+            except (OSError, KeyError):
+                # Cache missing or unreadable; fall back to source
+                pass
+        if use_cache:
+            # don't need to load from source file and reorganize by identity
+            return
+
+        # v8 specific loading: bounding boxes
+        with h5py.File(self._path, "r") as pose_h5:
+            ds = pose_h5.get("poseest/bbox")
```
keithshep, line 49: Recommend breaking this `with` block out as a method. Especially since there is an early exit before I feel like the flow of this can get a little confusing when it's all in `__init__`. Then invoke method `if not use_cache`

src/jabs/project/video_manager.py
```diff
@@ -75,11 +80,14 @@ def total_project_identities(self) -> int:
         """Get the total number of identities across all videos."""
         return self._total_project_identities
 
-    def load_video_labels(self, video_name) -> VideoLabels | None:
+    def load_video_labels(
+        self, video_name: Path, pose: PoseEstimation = None
+    ) -> VideoLabels | None:
```
keithshep, line 85: Maybe:

```suggestion
    def load_video_labels(
        self, video_name: Path | str, pose: PoseEstimation = None
    ) -> VideoLabels | None:
```

src/jabs/scripts/convert_parquet.py
```diff
@@ -181,12 +182,20 @@ def convert_data_frame(
                 )
                 jabs_confidences[frame, identity, jabs_keypoint.value] = 1.0
 
+        jabs_bboxes[frame, identity, 0] = [row["bb_left"], row["bb_top"]]
+        jabs_bboxes[frame, identity, 1] = [row["bb_right"], row["bb_bottom"]]
```
keithshep, line 186: I don't think you'll run into this but just as FYI older versions can use different col names. I found this snippet in murine repo:

```
        # Rename columns to match across different hydra output versions
        column_mapping = {'bbox_a': 'bb_left',
                  'bbox_b': 'bb_top',
                  'bbox_c': 'bb_right',
                  'bbox_d': 'bb_bottom',
                 }
        return df_parquet.rename(columns=column_mapping)
```

keithshep (approved):
nice addition

Issue #226, opened 2025-09-23
GUI crashing unexpectedly


emdavis02:
# GUI crashes (`zsh: killed jabs`) when opening certain folders

## Description  
I’m running into a reproducible crash when trying to open some data folders in the JABS GUI. The application opens fine, but as soon as I select the folder, the process dies with:
`zsh: killed     jabs`

## System details
- **OS:** macOS (Sonoma)  
- **RAM:** 24 GB  
- **Python:** tried with both 3.13.5 and 3.11.13  
- **Install method:**  
  ```bash
  python -m venv jabs.venv
  source jabs.venv/bin/activate
  pip install git+https://github.com/KumarLabJax/JABS-behavior-classifier.git

What I’ve tried
	•	The crash only happens for certain folders. Strangely, I can open a different folder with more data without any issues.
	•	The problematic folder contains paired .mp4 video(s) and corresponding .h5 pose estimation file(s).
	•	I checked the .h5 file with h5py, and it looks valid (keys: ['poseset', 'static_objects']).
	•	I am able to open the mp4, it is not corrupt
	•	I renamed files to a very simple schema (test.mp4, test_pose.h5), but the crash still happens.

Could you help me debug this issue?
Also, would it be possible to add more robust error handling/logging so the GUI reports the cause of the crash instead of exiting silently?
I'm happy to provide the files that trigger the crash if that helps.

gbeane:
next version of JABS will catch this error before crashing:

<img width="875" height="702" alt="Image" src="https://github.com/user-attachments/assets/ff217958-981c-4929-81c2-77e928ef2ac2" />

PR #230, opened 2025-10-03
add some more error checking in PoseEstimationV4

gbeane:
this addresses the issue described in this issue reported on GitHub:

https://github.com/KumarLabJax/JABS-behavior-classifier/issues/226

the pose file had incorrect values for instance_embed_id:

image

This was causing jabs to think there was a very large number of identities, which caused it to allocate excessively large numpy arrays and eventually get killed trying to read in the pose file.  

```bash
(JABS-behavior-classifier) gbeane@MLG-BH0689 JABS-behavior-classifier % jabs ~/Desktop/debug 
zsh: killed     jabs ~/Desktop/debug

```

This PR adds an extra check to make sure the values are in an expected range. If instance_embed_id values fall outside an expected range, we raise an exception which will cause the GUI to display an error message

### DIFF
--- a/src/jabs/pose_estimation/pose_est_v4.py
+++ b/src/jabs/pose_estimation/pose_est_v4.py
@@ -108,7 +108,6 @@ class PoseEstimationV4(PoseEstimation):
                     self._num_identities = 0
 
             # Validate instance_embed_id range: must be in [0, self._num_identities]
-            # use self._num_identities, which is previously set to the max value found in instance_embed_id
             if max_instance_id > self._num_identities:
                 raise PoseIdEmbeddingException(
                     f"Invalid instance_embed_id, values out of range: {file_path.name}"


keithshep (approved):

src/jabs/pose_estimation/pose_est_v4.py
```diff
@@ -95,6 +101,13 @@ def __init__(self, file_path: Path, cache_dir: Path | None = None, fps: int = 30
                 print(f"Warning: No identities found in pose file: {file_path}")
                 self._num_identities = 0
 
+            # Validate instance_embed_id range: must be in [0, self._num_identities]
+            # use self._num_identities, which is previously set to the max value found in instance_embed_id
+            if self._num_identities > instance_embed_id.shape[0]:
```
SkepticRaven, line 106: The shape of `poseest/instance_embed_id` is not necessarily limited to the number of mice. I think it's unlikely, but possible that we want 3 IDs assigned and only 2 are predicted at the same time.

`poseest/instance_id_center`'s first dimension should guarantee the number of assigned IDs.

PR #237, opened 2025-11-06
add jabs-cli prune command; implement project metadata merging for jabs-init command

gbeane:
adds a new command called "prune" to the `jabs-cli` script, 
provides a cli-based way for pruning unlabeled videos from a JABS project (functionality already exists in the GUI).

cli command is more complete in that it also removes derived files (GUI pruning tries to move videos, poses, and annotations to recycle bin, but does not remove cache files, features, predictions). GUI should be updated in a future PR to add missing functionality using new helper functions now available as part of this PR. 

Also updates behavior of importing metadata with `jabs-init`. Add option to merge or replace when importing new metadata file (previous behavior only allowed complete replacement, there was no way to merge)

### DIFF
--- a/src/jabs/scripts/cli.py
+++ b/src/jabs/scripts/cli.py
@@ -226,7 +226,10 @@ def prune(ctx, directory: Path, behavior: str | None):
                 click.echo(f"Warning: failed to delete derived file {file}: {e}")
 
         # remove from the project.json file
-        project.settings_manager.remove_video_from_project_file(video_paths.video_path.name)
+        project.settings_manager.remove_video_from_project_file(
+            video_paths.video_path.name, sync=False
+        )
+    project.settings_manager.save_project_file()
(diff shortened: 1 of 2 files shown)


src/jabs/scripts/cli.py
```diff
@@ -156,6 +156,82 @@ def rename_behavior(ctx, directory: Path, old_name: str, new_name: str) -> None:
...
+    for video_path in videos_to_prune:
+        click.echo(f" - {video_path.video_path.name}")
+
+    confirm = click.confirm("Do you want to proceed with pruning these videos?", default=False)
+
+    if not confirm:
+        click.echo("Pruning cancelled.")
+        return
+
+    for video_paths in videos_to_prune:
```
ptuan5: I would change this to `video_path` or `video`

PR #238, opened 2025-11-08
Speed up Project.get_labeled_features() with a persistent ProcessPoolExecutor with background pool warm-up on project load

gbeane:
### Summary

This PR introduces a persistent, instance-scoped, non-resizing ProcessPoolExecutor owned by each Project and a small QThread warm-up that pre-spawns worker processes when a project is loaded. The ProcessPoolExecutor is used to parallelize loading labeled frames in Project.get_labeled_features(), with parallelization implemented at the video-level (features can be loaded for multiple videos in parallel). 

Note: keeping a persistent pool that is warmed on project load results in a noticeably faster first run after opening a project (≈ ~6s saved when running training with warm executor pool vs cold)

### Motivation

Loading features for the labeled frames consumes a significant amount of the time during a JABS training cycle. The main motivation of this change is to parallelize the feature loading. This results in significant time savings even when features have been precomputed, but the speedup is even more dramatic when features need to be calculated rather than just loaded from disk.

### Key Changes

Project (backend)
- Persistent process pool
- Added an instance-level, non-resizing ProcessPoolExecutor. Currently this is used by get_labeled_features() but might be useful in other methods that will benefit from multi-process based parallelism.
- Pool size is fixed at construction via a new executor_workers: int | None parameter (defaults to os.cpu_count()).
- Created lazily on first use via _ensure_executor() (idempotent, create-once; no auto-resizing).
- Warm-up & lifecycle
  - warm_executor(wait: bool = True): Optional warm-up to pre-spawn workers (submits trivial jobs when wait=True).
  - shutdown_executor(): Explicit cleanup for CLI/app exit.
- Docs
  - Expanded Project class docstring with an “Executor pool” section (lifecycle, sizing, thread safety).

GUI (MainWindow)
- Warm on project load
  - Introduced MainWindow._WarmExecutorThread (QThread inner class) to call project.warm_executor() off the UI thread after a project loads.
  - Prevents UI stalls and avoids paying spawn/import cost on the first feature extraction/training run.
- Robustness when switching projects
  - Cancels/joins any in-flight warm thread on open/after load to avoid warming a stale Project.
  - Handles the “Internal C++ object already deleted” edge case by safely clearing stale thread refs and guarding .isRunning() calls.

Behavior / UX Impact
- From a user's perspective, training behavior is unchanged.
- No new UI elements
- Pool warm-up is silent and non-blocking.

### Performance
- Local timing showed reduction from approximately 13 seconds to ~4 seconds for a training cycle for a JABS project with 250 one-minute long videos with 33 of them containing labels.
- Submitting fewer jobs than workers is fine (idle workers).
- Submitting more jobs than workers queues tasks as expected.

### Thread Safety 
- The executor is instance-scoped (per Project) and created once.
- Submissions are thread-safe; training threads and warming threads can co-exist.
- Warm thread checks for interruption and verifies it’s still applicable to the current Project.
- When opening another project, any previous warm thread is interrupted and joined to avoid stale work.

### API Changes
- Project.__init__(..., executor_workers: int | None = None): optional fixed pool size.
- Project.warm_executor(wait: bool = True): pre-spawn worker processes (does not resize).
- Project.shutdown_executor(): explicit cleanup hook.

### Backward Compatibility
- Existing callers are unaffected.
- If you construct Project without executor_workers, the default remains sensible (os.cpu_count()).
- CLI usage should call shutdown_executor() at script end (good hygiene; avoids rare teardown warnings).

### Testing

- Manual tests:
  - Open project → warm thread starts, first extraction avoided cold-start latency. 
  - Rapidly switching projects: previous warm thread interrupted, no stale warmups; no crashes on deleted C++ objects.
  - Repeated runs confirm stable results and improved time-to-first-batch.

### Risks & Mitigations

- Zombie/stale threads: mitigated by interruption + wait + clearing references on finished/destroyed.
- Resource usage: persistent workers keep memory resident; mitigated by explicit shutdown_executor() on project close/app exit.
- Pool size choice: fixed at construction; future enhancement may add a guarded resize_executor() callable when idle.

### DIFF
--- a/src/jabs/project/parallel_workers.py
+++ b/src/jabs/project/parallel_workers.py
@@ -7,7 +7,7 @@ be executed by ProcessPoolExecutor workers, managed by Project.get_labeled_featu
 
 import json
 from pathlib import Path
-from typing import TypedDict
+from typing import TYPE_CHECKING, TypedDict
 
 import numpy as np
 import pandas as pd
@@ -18,6 +18,9 @@ from jabs.video_reader.utilities import get_fps
 
 from .track_labels import TrackLabels
 from .video_labels import VideoLabels
+
+if TYPE_CHECKING:
+    from jabs.pose_estimation import PoseEstimation
 
 
 class FeatureLoadJobSpec(TypedDict):
@@ -45,7 +48,7 @@ class CollectFeatureLoadResult(TypedDict):
     group_keys: list[tuple[str, int]]
 
 
-def _load_video_labels(annotations_path: Path, pose_est) -> VideoLabels | None:
+def _load_video_labels(annotations_path: Path, pose_est: "PoseEstimation") -> VideoLabels | None:
     """Load VideoLabels from a JSON file if present; else None."""
     ap = annotations_path
     if not ap.exists():


gbeane:
@SkepticRaven 

RAM usage will go up per worker. Each is a separate process with its own Python interpreter with imports loaded. It could be 100-300MB per worker extra memory overhead. 

I'll likely do this as a separate pull request (becauseI want to use a new "JABS Settings Dialog" that exists in another branch with an open pull request), but I'll add a control in the GUI for setting the number of workers and/or disabling the worker pool completely. If memory is tight, we can dial down the worker pool size, so trade less memory usage for longer run time.

PR #239, opened 2025-11-13
write out prediction files as they are generated

gbeane:
write out prediction files as they are generated instead of keeping in memory and writing them all out at the end of the prediction loop

When you clicked the "Classify" button in the JABS GUI it would predict behavior on all videos, holding results in memory. Then, once all videos have been processed, it would write out all of the prediction hdf5 files. 

This change has the JABS GUI write out the files as predictions are generated (as the classification thread loops over each video, we now write out the prediction file at the end of the loop)

### DIFF
--- a/src/jabs/project/project.py
+++ b/src/jabs/project/project.py
@@ -463,7 +463,7 @@ class Project:
             In the future, if the upstream caller were to provide compact arrays of length `len(indexes)`
             instead of full-length arrays, the copy logic would need to drop the indexing on the source side.
         """
-        # set up an output filename based on the behavior and video names
+        # set up an output filename based on the video names
         file_base = Path(video_name).with_suffix("").name + ".h5"
         output_path = self._paths.prediction_dir / file_base
 


src/jabs/project/prediction_manager.py
```diff
@@ -68,21 +69,21 @@ def write_predictions(
...
             poses: PoseEstimation object corresponding to the video.
             classifier: Classifier object used to generate predictions.
-            external_identities (list[int], optional): List mapping JABS identities to external identities.
 
         Returns:
             None
         """
         # TODO catch exceptions
-        with h5py.File(output_path, "a") as h5:
+        with h5py.File(output_path, "w") as h5:
```
SkepticRaven: I think this needs to be append mode. We write multiple behavior predictions to the same file (line 87/88 below). Opening in write mode will discard previous predictions.

SkepticRaven (changes requested):
Approve all except for the 1 comment.

PR #257, opened 2026-01-02
update README with pipx and uvx instructions

gbeane:
adds information about installing with pipx or running with uvx.

Also removes hard line wrapping from README.md to make file maintenance easier. JABS developers should enable soft line wrapping for .md files in their IDE settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,8 @@ Double-click these scripts in Windows Explorer to run them.
 
 The XGBoost Classifier has a dependency on the OpenMP library. This does not ship with macOS. XGBoost should work "out of the box" on other platforms. On macOS, you can install libomp with Homebrew (preferred) with the following command `brew install libomp`. You can also install libomp from source if you can't use Homebrew, but this is beyond the scope of this Readme.
 
+Because libomp is dynamically loaded by XGBoost it can be installed before or after installing jabs-behavior-classifier.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,10 +130,7 @@ Double-click these scripts in Windows Explorer to run them.
 
 ### Enabling XGBoost Classifier
 
-The XGBoost Classifier has a dependency on the OpenMP library. This does not ship with macOS. XGBoost should work "out 
-of the box" on other platforms. On macOS, you can install libomp with Homebrew (preferred) with the following 
-command `brew install libomp`. You can also install libomp from source if you can't use Homebrew, but this is beyond 
-the scope of this Readme.
+The XGBoost Classifier has a dependency on the OpenMP library. This does not ship with macOS. XGBoost should work "out of the box" on other platforms. On macOS, you can install libomp with Homebrew (preferred) with the following command `brew install libomp`. You can also install libomp from source if you can't use Homebrew, but this is beyond the scope of this Readme.
```
keithshep: Should we say something like "Because libomp is dynamically loaded by XGBoost it can be installed before or after installing jabs-behavior-classifier"

keithshep (approved):

README.md
gbeane: done.

PR #328, opened 2026-03-25
First draft of Cloud Platform Infrastructure ADR

bergsalex:
This PR includes a first draft proposal of our cloud infrastructure approach for the coming JABS work, based on recent in-person technology and priority discussions.


keithshep (approved):
The only thing that occurred to me is maybe we should be explicit about versioning in terms of whether it's in scope or not, and if it is what kinds of data should be versioned

bergsalex:
Moving this ADR to the jabs-hub repository.

PR #350, opened 2026-04-20
Add Parquet feature cache format and CLI options

gbeane:
This pull request adds support for Parquet as the default cache format, and enhance cache directory structure options. The changes clarify and expand documentation for feature cache formats and migration, add new CLI options for cache management.

For large projects, testing shows that by using the more efficient Parquet feature cache format, the classification loop in the GUI can see a ~40% speedup.

**Documentation improvements:**

* Expanded user and project setup guides to document both HDF5 and Parquet feature cache formats, migration steps, and cache directory structure, including new CLI options like `--use-pose-hash`, `--skip-window-cache`, and `--cache-format`. 

**Feature cache and CLI enhancements:**

* Added `--use-pose-hash` option to CLI tools and feature extraction to support more robust cache directory layouts and avoid filename collisions in shared caches. 
* Updated `jabs-init` to support explicit cache format selection (`--cache-format {hdf5,parquet}`).

**Core code and serialization updates:**

* Changed `CacheFormat` enum to inherit from `str` for easier JSON serialization and added `CACHE_FORMAT_KEY` constant for project settings. 
* Improved Parquet cache metadata serialization to ensure values are always stored as floats or None.

**Feature extraction and performance improvements:**

* Refactored feature extraction code to normalize operation settings, optimize per-frame feature handling, and add efficient flat feature access methods, reducing unnecessary data transformations. 
* Consolidated and simplified feature filtering logic for both per-frame and window features based on operation settings.

### DIFF
--- a/src/jabs/project/parallel_workers.py
+++ b/src/jabs/project/parallel_workers.py
@@ -6,6 +6,7 @@ be executed by ProcessPoolExecutor workers, managed by Project.get_labeled_featu
 """
 
 import json
+import logging
 import multiprocessing
 import sys
 from pathlib import Path
@@ -15,11 +16,14 @@ import numpy as np
 import pandas as pd
 
 import jabs.feature_extraction as fe
+from jabs.core.enums import CacheFormat
 from jabs.pose_estimation import open_pose_file
 from jabs.video_reader.utilities import get_fps
 
 from .track_labels import TrackLabels
 from .video_labels import VideoLabels
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from jabs.pose_estimation import PoseEstimation
@@ -40,6 +44,7 @@ class FeatureLoadJobSpec(TypedDict):
     cache_dir: Path | None
     behavior_settings: dict[str, object]
     behavior_name: str | None
+    cache_format: str
 
 
 class CollectFeatureLoadResult(TypedDict):
@@ -131,6 +136,14 @@ def collect_labeled_features(job: FeatureLoadJobSpec) -> CollectFeatureLoadResul
             continue
 
         # Feature extraction for this identity
+        try:
+            cache_format = CacheFormat(job["cache_format"])
+        except (KeyError, ValueError):
+            logger.error(
+                "Unknown cache_format %r in job spec; falling back to HDF5",
+                job.get("cache_format"),
+            )
+            cache_format = CacheFormat.HDF5
         features = fe.IdentityFeatures(
             video,
             identity,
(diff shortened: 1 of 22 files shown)


src/jabs/project/parallel_workers.py
```diff
@@ -131,18 +136,26 @@ def collect_labeled_features(job: FeatureLoadJobSpec) -> CollectFeatureLoadResul
...
 
         # Feature extraction for this identity
+        try:
+            cache_format = CacheFormat(job["cache_format"])
+        except (KeyError, ValueError):
+            logger.error(
+                "Unknown cache_format %r in job spec; falling back to HDF5",
+                job.get("cache_format"),
+            )
+            cache_format = CacheFormat.HDF5
```
keithshep: just double checking: should this exception not be fatal / fail-fast?

src/jabs/scripts/initialize_project.py
```diff
@@ -31,6 +35,14 @@ def generate_files_worker(params: dict):
...
     project = params["project"]
     pose_est = project.load_pose_est(project.video_manager.video_path(params["video"]))
+    try:
+        cache_format = CacheFormat(params.get("cache_format", CacheFormat.HDF5.value))
+    except ValueError:
+        logger.error(
+            "Unknown cache_format %r in job params; falling back to HDF5",
+            params.get("cache_format"),
+        )
+        cache_format = CacheFormat.HDF5
```
keithshep: candidate for a small utility function?

keithshep (approved):

bergsalex (approved):

src/jabs/project/parallel_workers.py
gbeane: maybe it should be fatal. I was thinking we can just fall back to the lowest common denominator, HDF5, but this shouldn't happen and if it did maybe it would be best to fail and have to "repair" the project by setting the "cache_format" value in the project.json file to a supported value

gbeane: so I think I'm going to just have the bare code here without wrapping it in try/except and let it raise a KeyError if "cache_format" is missing from he job spec or a ValueError if it's not a valid CacheFormat enum value

this shouldn't happen, and if it does we'll just propagate the exception back to the caller with the stack trace. The caller will catch the exception and re-raise a RunTimeError: "RuntimeError: Feature collection failed for video: <name>" with the original exception chained via __cause__

src/jabs/scripts/initialize_project.py
gbeane: I think this is unnecessary code -- the "cache_format" parameter is always set by the job producer AND it's value is going to be validated by Click

PR #354, opened 2026-04-22
Implement multi-class label visualization and refactor timeline widgets

gbeane:
This pull request introduces support for multi-class label visualization in the timeline and label overview widgets, enabling both binary and multi-class behavior labeling with direct color palette mapping. The changes focus on normalizing label and prediction data to color lookup table (LUT) indices, providing extensible color palette generation, and refactoring widget APIs to accept pre-normalized arrays for flexible rendering.

**Multi-class label and color palette support:**

* Added `VideoLabels.build_multiclass_label_array` to generate per-frame class-index arrays for multi-class behavior visualization, mapping each frame to a unique color LUT index for direct lookup.
* Introduced `make_behavior_color_map` and `build_multiclass_color_lut` in `colors.py` to generate visually distinct per-behavior color palettes, excluding fixed UI colors to avoid clashes, and returning RGBA LUTs for use in widgets.

**Widget API and rendering refactor:**

* Refactored label and prediction widget APIs (`ManualLabelWidget`, `LabelOverviewWidget`, etc.) to accept pre-normalized LUT-index arrays (`np.int16`), rather than raw `TrackLabels` or prediction values. Added methods to set custom color LUTs for multi-class mode. 
* Updated painting logic in widgets to use the provided LUT indices and color tables directly, ensuring correct color mapping for both binary and multi-class scenarios. 

**Normalization utilities and API surface:**

* Added utility functions `track_labels_to_lut_indices` and `binary_predictions_to_lut_indices` for converting binary labels and predictions to LUT indices, centralizing normalization logic and ensuring consistency across the UI.
* Updated imports and method calls throughout the UI to use these normalization utilities, ensuring all label and prediction data is properly prepared before being passed to widgets. 

These changes lay the groundwork for robust multi-class behavior visualization and simplify future extensions to the label and prediction rendering pipeline.


src/jabs/ui/colors.py
```diff
@@ -30,3 +32,96 @@
...
+        exclude_colors=_EXCLUDED_BEHAVIOR_PALETTE_COLORS,
+        rng=0,
+    )
+    return {
+        name: QColor(int(r * 255), int(g * 255), int(b * 255))
+        for name, (r, g, b) in zip(behavior_names, rgb_floats, strict=True)
+    }
+
+
+def build_multiclass_color_lut(
```
bergsalex, line 83: I don't think this function is used anywhere but the tests. Ignore if this is intentional set-up for future usage.

src/jabs/project/video_labels.py
```diff
@@ -93,6 +96,53 @@ def iter_identity_behavior_labels(self):
             for behavior, track_labels in behaviors.items():
                 yield identity, behavior, track_labels
 
+    def build_multiclass_label_array(self, identity: str, behavior_names: list[str]) -> np.ndarray:
```
bergsalex, line 99: I don't think this function is used anywhere but the tests. Ignore if this is intentional set-up for future usage.

bergsalex (approved):

src/jabs/ui/stacked_timeline_widget/label_overview_widget/label_overview_widget.py
```diff
@@ -118,12 +118,28 @@ def framerate(self, fps: int) -> None:
...
+        """Set the color lookup table for both child widgets.
+
+        Propagates a custom LUT to the timeline and label widgets, enabling
+        multi-class color rendering.
+
+        Args:
+            lut: RGBA array of shape ``(N, 4)`` mapping class indices to colors.
+        """
+        self._timeline_widget.set_color_lut(lut)
+        self._label_widget.set_color_lut(lut)
```
keithshep, line 136: is it worth validating lut shape with ValueError on fail?

src/jabs/ui/stacked_timeline_widget/label_overview_widget/manual_label_widget.py
```diff
@@ -92,6 +95,20 @@ def __init__(self, *args, **kwargs) -> None:
...
+    def set_color_lut(self, lut: npt.NDArray[np.uint8]) -> None:
+        """Replace the color lookup table used to render label frames.
+
+        In binary mode this is never called and the class-level ``COLOR_LUT``
+        is used.  In multi-class mode ``StackedTimelineWidget`` calls this with
+        the per-behavior palette produced by
+        :func:`jabs.ui.colors.build_multiclass_color_lut`.
+
+        Args:
+            lut: RGBA array of shape ``(N, 4)`` mapping class indices to colors.
```
keithshep, line 107: same as prev

src/jabs/ui/stacked_timeline_widget/label_overview_widget/timeline_label_widget.py
```diff
@@ -25,6 +27,54 @@
...
+    binned = padded.reshape(size, bin_size)  # (size, bin_size)
+
+    # Count occurrences of each class per bin: shape (size, n_classes)
+    counts = np.zeros((size, n_classes), dtype=np.float32)
+    for c in range(n_classes):
+        counts[:, c] = (binned == c).sum(axis=1)
+
+    # Proportional weighted average color per bin
+    colors = (counts @ lut.astype(np.float32)) / bin_size
+    return colors.clip(0, 255).astype(np.uint8)
```
keithshep: non-blocking. Ignore this if it's too minor a concern.

I think color mixing can get weird with RGB. ChatGPT tells me:

> Most RGB values in UI code are effectively sRGB, which is gamma-encoded.
Averaging gamma-encoded RGB channels does not correspond to averaging light.
>
> So if you want “correct” mixing of emitted light, the usual recipe is:
>
> 1. convert sRGB to linear RGB
> 2. average in linear RGB
> 3. convert back to sRGB

Also if the alpha is a measure of confidence (no idea if it is) then the color mixing should be weighted by alpha rather than them sharing the alpha result (I think)

keithshep (approved):

src/jabs/ui/stacked_timeline_widget/label_overview_widget/timeline_label_widget.py
gbeane: the "zoomed out" label and prediction overview bars that show the entire video do not use the alpha channel to encode confidence.  This is for the skinny bars on top that show an overview of the whole video:

image

In this case the label or prediction classes are downsampled since there might be more frames than the widget's width in pixels. During downsampling, each bin is assigned a color (originally this was: bin all "not-behavior" = blue, bin all "behavior" = "orange", bin is a mix of behavior/not-behavior = a fixed purple color)

the new code attempts to have each class contribute proportionally to the bin's color


I think I might go ahead with this color mixing but keep your comment in mind

gbeane: actually, I implemented the change to blend in linear light space, and also a change to not have background color contribute to the bin color (unless the bin is all background)

PR #360, opened 2026-04-25
Implement MultiClassClassifier with validation and visualization enhancements

gbeane:
Sorry for the monster PR.  This also merges upstream changes from main since I needed the new "compact" mode for `ManualLabelWidget` and `PredictedLabelWidget`. I probably should have done a "merge main into feature/multiclass" pull request first so that didn't get mixed with my changes. 


## Description of new functionality that wasn't merged from main
This PR extends StackedTimelineWidget (stacked_timeline_widget.py) and its surrounding infrastructure to support multiclass classifier visualization. In multiclass mode the widget builds an entirely different per-identity layout: a combined argmax bar (PredictionOverviewWidget) that shows the winning class color across all frames, followed by one per-class row per behavior (PerClassPredictionOverviewWidget). The per-class rows use the new BehaviorProbabilityWidget (behavior_probability_widget.py) — a subclass of PredictedLabelWidget that overrides _build_frame_colors to render the behavior color at alpha = probability rather than toggling a hard "not-this-class" gray. A _BehaviorLegendWidget strip at the top maps colors to behavior names.  PredictedLabelWidget was refactored to extract _build_frame_colors as an overridable hook to make that subclassing clean; reviewers should check that the base behavior is unchanged. The multiclass LUT and color map are built via build_multiclass_color_lut and make_behavior_color_map in colors.py and propagated down to each sub-widget via set_color_lut.                                                                                                      
                                                                                                                                                             
The wiring into the main window lives in central_widget.py, which detects classifier mode and routes prediction arrays to the appropriate widget structure  — it also explicitly blocks binary prediction data from flowing into the multiclass layout (and vice versa). Four new checkable Timeline menu actions  (menu_builder.py / menu_handlers.py) let users collapse or hide inactive-identity rows in all-animals mode; they are disabled until both multiclass mode and all-animals identity mode are simultaneously active. One additional fix worth noting: TimelineLabelWidget._update_bar previously produced no pixmap when _labels was None, causing the overview bar to be invisible before predictions were first run in multiclass mode; it now renders a solid background-color bar in that state to match binary mode's behavior. (in binary mode we got around this by generating a numpy array with no predictions to pass in so a blank bar was displayed -- now the widget handles this itself for multi-class)

As I was working on integrating the multi-class view options into the app, I found several things that needed to be changed/fixed (even though a previous PR tried to lay the foundation for multi-class in the StackedTimelineWidget, there are some things I missed or decided to go a different direction after seeing the result in practice)




I also added a script `dev/preview_multi_class_timeline.py` that lets you play with this and the various view options I added using fake data. 

For multi-class, all animals mode, there are 4 different view options that you can toggle on/off independently

Demo script showing single animal view:
<img width="1402" height="690" alt="Screenshot 2026-04-24 at 8 16 44 PM" src="https://github.com/user-attachments/assets/de05aceb-d766-4839-856a-45d30ffdaddc" />

bars are
1. label overview
2. labels detailed view
3. predictions (combined) overview
4. predictions (combined) detailed view

then prediction overview and details for each class separately 


----

shot of it being used in JABS. Showing all animals, but with inactive animals set to only show the label and combined prediction overviews (not detail bars, no per-class bars). 
<img width="1800" height="1441" alt="Screenshot 2026-04-24 at 9 04 36 PM" src="https://github.com/user-attachments/assets/2bb7bbae-3fa1-4d34-b73b-4c85782bdadf" />


I would recommend playing with the demo app. 

Note: this has only been tested in JABS for **labeling** in multi-class mode. Viewing actual predictions hasn't been tested yet (only simulated values in the demo app).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,6 +158,8 @@ cache invalidation logic (version mismatch, pose hash, distance scale).
 - Use American English spelling (e.g., "behavior" not "behaviour").
 - Prefer importing at the top of the module. It is acceptable to import within
   functions/methods if an expensive import is only needed in specific code paths.
+- Do not use ambiguous `–` (EN DASH) in docstrings. Use `-` (HYPHEN-MINUS) to avoid 
+  Ruff RUF001 error.
 
 ## Coding Standards
 
--- a/dev/preview_multiclass_timeline.py
+++ b/dev/preview_multiclass_timeline.py
@@ -1,0 +1,349 @@
+"""Preview script: StackedTimelineWidget with fake multi-class data.
+
+Run with:
+    uv run python preview_multiclass_timeline.py
+
+Displays 3 identities, 3 behaviors, ~2 000 frames.  A slider scrubs through
+frames; radio buttons toggle binary / multi-class mode and single / all-identity
+view so you can compare layouts.
+"""
+
+import sys
+
+import numpy as np
+from PySide6.QtCore import Qt
+from PySide6.QtGui import QKeySequence, QShortcut
+from PySide6.QtWidgets import (
+    QApplication,
+    QButtonGroup,
+    QCheckBox,
+    QGroupBox,
+    QHBoxLayout,
+    QLabel,
+    QMainWindow,
+    QPushButton,
+    QRadioButton,
+    QScrollArea,
+    QSlider,
+    QVBoxLayout,
+    QWidget,
+)
+
+from jabs.core.enums import ClassifierMode
+from jabs.ui.stacked_timeline_widget import StackedTimelineWidget
+
+# ---------------------------------------------------------------------------
+# Fake pose shim (duck-types the attributes/method that _reset_layout needs)
+# ---------------------------------------------------------------------------
+
+
+class _FakePose:
+    def __init__(self, num_identities: int, num_frames: int) -> None:
+        self.num_identities = num_identities
+        self.num_frames = num_frames
+
+    def identity_index_to_display(self, index: int) -> str:
+        return f"Mouse {index + 1}"
+
+
+# ---------------------------------------------------------------------------
+# Fake data generation
+# ---------------------------------------------------------------------------
+
+NUM_IDENTITIES = 3
+NUM_FRAMES = 2000
+FRAMERATE = 30
+BEHAVIORS = ["grooming", "rearing", "locomotion"]
+
+rng = np.random.default_rng(seed=42)
+
+
+def _make_bouts(n_frames: int, avg_gap: int = 120, avg_dur: int = 60) -> np.ndarray:
+    """Return a boolean mask with random on/off bouts."""
+    mask = np.zeros(n_frames, dtype=bool)
+    frame = 0
+    while frame < n_frames:
+        frame += max(10, int(rng.exponential(avg_gap)))
+        dur = max(5, int(rng.exponential(avg_dur)))
+        mask[frame : frame + dur] = True
+        frame += dur
+    return mask
+
+
+def _make_multiclass_labels(n_frames: int) -> np.ndarray:
+    """Build a combined class-index label array for one identity.
+
+    Index layout (matches build_multiclass_color_lut):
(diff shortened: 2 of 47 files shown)


src/jabs/ui/main_window/central_widget.py
```diff
@@ -724,15 +794,27 @@ def _set_label_track(self) -> None:
...
+                )
+            else:
+                label_list = self._get_label_list()
+                self._stacked_timeline.set_labels(
+                    [track_labels_to_lut_indices(t) for t in label_list],
+                    mask_list,
+                )
+                if self._label_overlay_mode == PlayerWidget.LabelOverlayMode.LABEL:
+                    # if configured to show labels, update the player widget with the new labels
+                    self._player_widget.set_labels([labels.get_labels() for labels in label_list])
```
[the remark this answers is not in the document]
gbeane: updated to set the current behavior's TrackLabels, but there will be a follow up ticket for making the label and prediction overlay multi-class aware. It currently does not work as desired, but will be addressed in its own ticket.

dev/preview_multiclass_timeline.py
```diff
@@ -0,0 +1,381 @@
...
+        timeline_h = self._timeline.sizeHint().height()
+        chrome = self.frameGeometry().height() - self.geometry().height()
+        needed = (
+            controls_h
+            + timeline_h
+            + margins.top()
+            + margins.bottom()
+            + chrome
+            + 8  # a little breathing room
+        )
```
keithshep, line 363: lol @ the breathing room comment

src/jabs/classifier/factories.py
```diff
@@ -36,6 +58,28 @@ def make_catboost(n_jobs: int, random_seed: int | None) -> CatBoostClassifier:
...
+
+    Args:
+        n_jobs: Number of parallel jobs.
+        random_seed: Random seed for reproducibility.
+
+    Returns:
+        CatBoostClassifier configured for multi-class classification.
+    """
+    return CatBoostClassifier(
+        loss_function="MultiClass",
```
keithshep, line 75: should we use enum for loss function?

src/jabs/project/settings_manager.py
```diff
@@ -88,6 +89,21 @@ def project_metadata(self) -> dict:
...
+            ClassifierMode: The configured classifier mode, defaulting to BINARY.
+        """
+        mode_str = self._project_info.get("settings", {}).get(
+            CLASSIFIER_MODE_KEY, DEFAULT_CLASSIFIER_MODE.value
+        )
+        try:
+            return ClassifierMode(mode_str)
+        except ValueError:
+            return DEFAULT_CLASSIFIER_MODE
+
```
keithshep, line 110: should we try to recover from this exception or should it raise a user visible error?

CLAUDE.md
```diff
@@ -155,9 +155,11 @@ cache invalidation logic (version mismatch, pose hash, distance scale).
 - `__init__.py` files: unused imports (F401) are allowed
 - **Naming**: `PascalCase` classes, `snake_case` functions/methods/files,
   `UPPER_SNAKE_CASE` constants, `_` prefix for private members
-- Prefer American English spelling (e.g., "behavior" not "behaviour").
+- Use American English spelling (e.g., "behavior" not "behaviour").
 - Prefer importing at the top of the module. It is acceptable to import within
   functions/methods if an expensive import is only needed in specific code paths.
+- Do not use ambiguous `–` (EN DASH) in docstrings. Use `-` (HYPHEN-MINUS) to avoid 
+  Ruff RUF001 error.
```
keithshep, line 162: I've also noticed use of unicode arrow instead of "->" in my comments and doc strings. Don't know if there is a more general instruction we can add like "avoid unicode glyphs such as arrows ..."

There's nothing for you to fix here. this just made me think

keithshep (approved):

CLAUDE.md
gbeane: I don't mind the arrows as much (but it does scream "an LLM wrote this docstring")

The EN DASH thing is super annoying because our Ruff config treats it as an error

PR #365, opened 2026-04-28
Check for conflicting labels in binary to multiclass migration

gbeane:
This pull request introduces a robust validation mechanism to ensure that, when switching a project to multi-class classification mode, no video contains overlapping behavior labels for the same identity and frame. The changes include both backend logic for detecting these overlaps and UI enhancements to warn users and block mode changes if conflicts exist. Comprehensive tests are also added to guarantee correct behavior.

A future enhancement could be to introduce a graphical conflict resolution strategy, but that is out of scope for this pull request. Instead, mode switching is blocked and the user is presented with a list of videos that contain conflicting behavior labels. 

**Validation and Conflict Detection for Multi-class Mode:**

* Added a `get_overlapping_behavior_label_videos` method to the `Project` class to scan all videos and identify those with frames labeled with multiple behaviors for the same identity. Returns a sorted list of conflicting video filenames.
* Introduced a background thread (`_OverlapCheckThread`) in the settings dialog to perform the overlap check asynchronously, keeping the UI responsive.

**User Interface Enhancements:**

* Updated the `ProjectSettingsDialog` to use the new overlap check when switching to multi-class mode. If conflicts are found, a warning dialog blocks the mode change and lists the problematic videos; otherwise, the user is warned about label preservation before proceeding.
* Modified the settings dialog instantiation to pass the `Project` object, enabling access to project-level validation methods. 

**Testing:**

* Added a suite of unit tests for `get_overlapping_behavior_label_videos`, covering edge cases (no videos, no labels, no conflicts, conflicts, multiple videos, and special handling for "None" behavior).

**Dependency and Import Updates:**

* Updated imports in relevant modules to support new functionality and dependencies, including constants, enums, and dialog components.

### DIFF
--- a/src/jabs/project/project.py
+++ b/src/jabs/project/project.py
@@ -16,7 +16,7 @@ import numpy as np
 import pandas as pd
 
 import jabs.feature_extraction as fe
-from jabs.core.constants import CACHE_FORMAT_KEY
+from jabs.core.constants import CACHE_FORMAT_KEY, MULTICLASS_NONE_BEHAVIOR
 from jabs.core.enums import CacheFormat, CrossValidationGroupingStrategy, ProjectDistanceUnit
 from jabs.pose_estimation import PoseEstimation, open_pose_file
 
@@ -27,6 +27,7 @@ from .project_paths import ProjectPaths
 from .project_utils import to_safe_name
 from .session_tracker import SessionTracker
 from .settings_manager import SettingsManager
+from .track_labels import TrackLabels
 from .video_labels import VideoLabels
 from .video_manager import VideoManager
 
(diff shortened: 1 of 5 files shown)


src/jabs/project/project.py
```diff
@@ -526,6 +527,37 @@ def save_predictions(
...
+        simultaneously, including the reserved "None" behavior.
+
+        Returns:
+            Sorted list of video filenames containing at least one overlap.
+            An empty list means no conflicts exist.
+        """
+        conflicting: list[str] = []
+        for video in self._video_manager.videos:
+            pose = self.load_pose_est(self._video_manager.video_path(video))
+            labels = self._video_manager.load_video_labels(video, pose)
```
keithshep: is conflict being checked for the fragmented labels or unfragmented? I would guess we want unfragmented but not 100% sure about that

keithshep (approved):

src/jabs/project/project.py
gbeane: unfragmented

PR #366, opened 2026-04-29
parallelize project load

gbeane:
### Problem

Opening a project in the GUI performed O(N) sequential file I/O operations for every video in the
project — three independent serial loops, each opening the same HDF5 pose files:

**Loop 1 — `VideoManager._validate_video_frame_counts()`**
Called on every GUI project open (`enable_video_check=True`). For each video, it opened the pose
HDF5 file to read `poseest/points.shape[0]` and also opened the video file via OpenCV to read the
frame count. These N pairs of file opens happened one at a time, blocking project load for the full
duration.

**Loop 2 — `FeatureManager.__initialize_pose_data()`**
Also ran on every open. For each video, it opened the pose HDF5 file *2–3 additional times* in
separate `with h5py.File(...)` calls:
- `get_static_objects_in_file()` — one open
- `get_points_per_lixit()` — a second open (if lixit present)
- `PoseEstimation.get_pose_file_attributes()` — a third open for `cm_per_pixel`

**Loop 3 — `VideoManager._load_video_metadata()`**
On first open, this created a full `PoseEstimation` object per video to read `num_identities`.
That constructor loads the entire keypoints array into memory (all frames × all keypoints × 2
coordinates), then discards the object after reading one number. For V6 files this also loads all
segmentation data. After first open, the result was cached under
`video_files[video]["identities"]` in `project.json` — but the FeatureManager loops still ran
regardless.

**The stale-cache problem with `video_files.identities`**
The `project.json` caching of identity counts was also fragile: if a pose file was re-exported
(e.g. ghost detections cleaned up), the identity count could change but the stale cached value
would continue to be used on subsequent loads.

---

### Solution

**One parallel scan pass replaces all three serial loops.**

A new module-level worker function `scan_video_metadata()` in `parallel_workers.py` opens each
HDF5 file exactly once and returns everything both `VideoManager` and `FeatureManager` need: frame
count, identity count, static objects, lixit keypoints, and whether `cm_per_pixel` is present. If
`enable_video_check=True`, it also reads the video frame count in the same call.

`Project.__init__()` runs this scan via the application-scoped `ProcessPoolManager` (up to 6
workers, already warm at startup) immediately after `SettingsManager` init. Results are collected
into a `dict[str, VideoScanResult]` and injected into both `VideoManager` and `FeatureManager`,
which skip their per-video I/O entirely when pre-loaded data is available.

When `process_pool` is `None` (CLI scripts), the same workers run sequentially — no functional
difference from before, but still 3–5× fewer HDF5 opens per video.

**Identity count without loading keypoints.**
The new `_get_identity_count()` helper reads identity count from an already-open `h5py.File`
using shape-attribute reads only:
- V2: returns 1 (from filename, no HDF5 access at all)
- V3: `poseest/points.shape[1]` — shape attribute only
- V4–V8: `poseest/instance_id_center.shape[0]` if the dataset exists — shape attribute only
- V4–V8 fallback: max of `poseest/instance_embed_id[id_mask==0]` — loads a small integer array, not the full keypoints

No `PoseEstimation` object is created. No keypoints are loaded into memory during project init.

**`video_files.identities` caching removed.**
Since identity count is now always read from the HDF5 on every open (fast, combined with the
existing scan), there is no need to cache it. This eliminates the stale-cache risk. Existing
`project.json` files that still contain the key are unaffected — the value is simply ignored.

**`SettingsManager.video_metadata()` bug fixed.**
This method previously did `self._project_info["video_files"][video]` — a hard dict access that
raises `KeyError` if the `video_files` section is absent, contradicting its own docstring which
says "empty dict if none exists". Since we no longer guarantee `video_files` is written at
project-init time, this bug was exposed and fixed: the method now uses `.get()` at every level.

**`PoseEstimationV8` cleanup.**
`_load_from_h5()` was redundantly re-deriving `_num_identities` from `instance_embed_id` despite
`super().__init__()` (V4's init) having already computed it correctly via `instance_id_center`.
Verified across 139 real V8 files that both methods always agree. The redundant block was removed,
making V8 consistent with V4–V7.

---

#### 1. `_get_identity_count()` — version-specific identity count logic

This is the highest-risk new code. It replicates logic that lives across several `PoseEstimation`
subclasses without using those classes. 

#### 2. `scan_video_metadata()` — pickling and worker safety

This function runs in a forked/spawned subprocess. Reviewers should confirm:

- No references to non-picklable state (class instances, open file handles, closures). All arguments are paths, primitives, or ints.
- `h5py.File` is opened fresh inside the worker (not inherited from the parent process). HDF5 is safe to use in forked workers as long as no file handle is shared across the fork boundary.
- The macOS fork/LAPACK guard from `collect_labeled_features` is intentionally omitted here — `scan_video_metadata` uses only h5py and OpenCV, neither of which has the Accelerate/LAPACK fork issue. If that assessment is wrong, the guard should be added.

#### 3. `SettingsManager.video_metadata()` behavioral change

The old implementation raised `KeyError` for any project that had never had video metadata saved.
The fix changes it to return `{}` in all missing-key cases. The method's docstring already
documented the `{}` return, so this is a bug fix — but reviewers should confirm no caller depends
on the `KeyError` being raised (e.g. to detect whether a video exists in the project):

### DIFF
--- a/src/jabs/pose_estimation/pose_est_v8.py
+++ b/src/jabs/pose_estimation/pose_est_v8.py
@@ -59,22 +59,11 @@ class PoseEstimationV8(PoseEstimationV7):
 
             bboxes = ds[:]
 
-            # Load identity mapping arrays, needed to reorganize bboxes by identity
+            # Load identity mapping arrays, needed to reorganize bboxes by identity.
+            # _num_identities was already set correctly by PoseEstimationV4.__init__
+            # via super().__init__(); no need to re-derive it here.
             instance_embed_id = pose_h5["poseest/instance_embed_id"][:]
             id_mask = pose_h5["poseest/id_mask"][:]
-
-            # Determine number of identities the same way v4 does: mask out invalids and take max
-            if instance_embed_id.shape[1] > 0:
-                valid = id_mask == 0
-                if valid.any():
-                    # instance_embed_id is 1-based; take max over valid entries
-                    self._num_identities = int(instance_embed_id[valid].max())
-                else:
-                    print(f"Warning: All identities masked in pose file: {self._path}")
-                    self._num_identities = 0
-            else:
-                print(f"Warning: No identities found in pose file: {self._path}")
-                self._num_identities = 0
 
             # Prepare an array grouped by identity, matching the v4 keypoint transform logic.
             # Shapes:
--- a/src/jabs/project/feature_manager.py
+++ b/src/jabs/project/feature_manager.py
@@ -1,3 +1,5 @@
+from typing import TYPE_CHECKING
+
 import jabs.feature_extraction as feature_extraction
 from jabs.core.enums import ProjectDistanceUnit
 from jabs.pose_estimation import (
@@ -10,6 +12,9 @@ from jabs.pose_estimation import (
 
 from .project_paths import ProjectPaths
 from .video_manager import VideoManager
+
+if TYPE_CHECKING:
+    from .parallel_workers import VideoScanResult
 
 
 class FeatureManager:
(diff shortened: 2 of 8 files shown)


keithshep (approved):
looks good

PR #382, opened 2026-05-19
normalize feature cache subdir name; auto-rename legacy dirs

gbeane:
## Summary

- `jabs-classify` and `jabs-cli compute-features` wrote feature caches under `<feature_dir>/<name>_pose_est_vN/<id>/`  (because it used the pose file stem as the subdirectory name) while `jabs-init` and the GUI used `<feature_dir>/<name>/<id>/` (because it used the video file stem as the subdirectory name). Same pose file, two different cache layouts.
- Added `pose_file_stem()` to `jabs.core.utils` and used it inside `IdentityFeatures.__init__` so all callers land on the bare-stem layout regardless of whether they pass a pose file or a video file.
- `IdentityFeatures.__init__` now also performs a one-shot rename of legacy `<name>_pose_est_vN/` dirs to `<name>/` so caches computed by older versions remain discoverable. The rename is skipped (with the legacy dir left alone) when the normalized destination already exists, to avoid collisions between sibling pose files that share a stem.
- Refactored `scripts/classify.py` to drop its local `get_pose_stem` regex helper in favor of the shared utility.

### DIFF
--- a/packages/jabs-core/src/jabs/core/utils/utilities.py
+++ b/packages/jabs-core/src/jabs/core/utils/utilities.py
@@ -65,6 +65,27 @@ def get_bool_env_var(var_name, default_value=False) -> bool:
     return value.lower() in ("true", "1", "yes", "on", "y", "t")
 
 
+_POSE_SUFFIX_RE = re.compile(r"_pose_est_v\d+$")
+
+
+def pose_file_stem(path: str | Path) -> str:
+    """Return the base name of a pose file with the ``_pose_est_vN`` suffix removed.
+
+    For example, ``"video_pose_est_v6.h5"`` becomes ``"video"``. If the input does
+    not include the ``_pose_est_vN`` suffix, the stem is returned unchanged
+    (this allows callers to pass either a pose file path or a video file path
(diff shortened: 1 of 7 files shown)


packages/jabs-core/src/jabs/core/utils/utilities.py
```diff
@@ -65,6 +65,27 @@ def get_bool_env_var(var_name, default_value=False) -> bool:
     return value.lower() in ("true", "1", "yes", "on", "y", "t")
 
 
+_POSE_SUFFIX_RE = re.compile(r"_pose_est_v\d+$")
```
keithshep, line 68: I think the `$` means it won't match if string ends with ".h5" (I could be wrong). Is that intended?

keithshep (approved):
seems like a win to me

packages/jabs-core/src/jabs/core/utils/utilities.py
gbeane: the .h5 is getting removed by Path().stem:

`return _POSE_SUFFIX_RE.sub("", Path(path).stem)`

PR #384, opened 2026-05-27
add jabs-cli update-labels subcommand

keithshep:
## Summary

Adds a new `jabs-cli update-labels` subcommand: the **inverse** of `jabs-cli update-pose`. Where `update-pose` keeps a project's labels and replaces its pose, `update-labels` keeps the target project's pose and replaces its labels with labels imported from another JABS project, remapping identities via bbox-IoU matching.

The shared label-remap machinery in `update_pose.py` is parameterized so both commands use it without duplication. A pre-existing bug in failure-annotation tag length (over `MAX_TAG_LEN=32`) is fixed at the same time.

## What's new

- **`jabs-cli update-labels <target> <source>`** — copies labels from another JABS project into the target, remapping identities via the same IoU algorithm used by `update-pose`. Per-video replace semantics (videos the source doesn't cover are left untouched). Behaviors not present in the target's `project.json` are merged in from the source so imported labels are usable in the GUI. Predictions are cleared (stale relative to new labels); pose, features, classifiers, and the performance cache are preserved.
- **Auto-scaffold** a minimal JABS project when the target is a bare `videos + pose` directory with no `jabs/` subdirectory. The scaffold creates the standard subdirectory layout and an empty `project.json`; features are **not** generated (run `jabs-init` separately afterwards). Source validation runs before scaffolding so a bad source path can't leave an orphaned `jabs/`.
- **Backup + manual-restore** behavior, parallel to `update-pose`: timestamped zip under `<target>/.backup/`, manual cleanup + restore instructions printed on failure.

## How it works

- Two disposable staging projects (`source_stage`, `dest_stage`) are constructed; the source's labels are remapped block-by-block against the target's existing pose via median bbox IoU.
- The shared `_run_staged_label_remap` / `_remap_labels_for_video` helpers (in `update_pose.py`) gain optional parameters: `videos` (preflighted video list — passed by both commands so iteration only touches videos validated by preflight) and `failure_description_phrase` (distinguishes "pose update" vs "label update" in failure annotation descriptions).
- Behavior merge is additive: source-only behaviors are added to the target's `project.json`; existing target behavior configs are preserved.

## Notable fixes / refactors

- **Failure-annotation tag length:** `update-pose` was producing tags of 33–37 chars, exceeding `TimelineAnnotations.MAX_TAG_LEN = 32`. Tags are silently dropped on load, so `--annotate-failures` was effectively broken even before this branch. Tags are now `behavior-remap-failed` (21) and `not-behavior-remap-failed` (25) — shared by both commands; the description text identifies which operation produced the annotation.
- **`update_pose.py` helpers parameterized for reuse:** `_create_backup_archive` (`include_pose_files`, `prefix`), `_validate_live_update_targets` (`require_writable_pose_files`, `derived_dir_names`), `_restore_cleanup_paths` (`replacement_pose_files` may be `None`).
- **Preflight reordered** so source validity is checked before any target-side mutation (scaffold).

## Documentation

- New `## jabs-cli update-labels` section in `cli-tools.md` (both `docs/user-guide/` and `src/jabs/resources/docs/user_guide/`), with explicit "inverse of update-pose" framing and bidirectional cross-links.
- Command summary table updated.
- In-app User Guide tree (`user_guide_dialog.py`) gains expandable `jabs-cli` node listing `update-pose` and `update-labels` so both are discoverable from the GUI sidebar.


src/jabs/scripts/cli/update_labels.py
```diff
@@ -0,0 +1,533 @@
...
+import tempfile
+from pathlib import Path
+
+import click
+
+from jabs.pose_estimation import get_pose_path
+from jabs.project import Project
+
+from .update_pose import (
+    _copy_file_atomic,
```
gbeane: Maybe `_copy_file_atomic` should get moved from cli.update_pose to jabs.core.utils -- it seems like something that could be generally useful.

gbeane (approved):

gbeane:
looks good to me

PR #395, opened 2026-06-09
Determine closest lixit from mouse centroid instead of nose keypoint

gbeane:
## Summary

In multi-lixit arenas, JABS selects the closest lixit per frame to compute the lixit distance and angle features. Previously this selection used the distance from the **nose** keypoint to each lixit tip. When the nose dropped out of the pose, the choice became unstable and could flip between lixit.

This branch switches the closest-lixit selection to use the **mouse centroid** (convex hull center) instead. The centroid is robust to dropout of any individual keypoint, keeping the choice stable across frames.

## Changes

- `LixitDistanceInfo` now picks the closest lixit using centroid-to-tip distance rather than nose-to-tip distance.
- Added per-frame centroid caching in `LixitDistanceInfo`, shared with `MouseLixitAngle` so the convex-hull centroid is computed only once (removes the duplicate centroid computation that previously lived in `MouseLixitAngle.per_frame`).
- Added `_closest_lixit_per_frame`, which fills frames with no valid centroid (too few keypoints to form a convex hull) from the temporally nearest valid frame (ties prefer the earlier frame); falls back to lixit 0 if no frame has a valid centroid.
- Bumped `FEATURE_VERSION` 16 → 17 to invalidate cached features.
- Updated user-guide feature docs (both `docs/user-guide/features.md` and the in-app copy).

## Tests

Added tests in `tests/feature_extraction/landmark_features/test_lixit_distance.py` covering the centroid-based selection and the nearest-valid-frame gap filling, plus a `fresh_pose_v5` fixture that installs custom lixit positions without mutating the shared module-scoped pose fixture.

### DIFF
--- a/src/jabs/feature_extraction/features.py
+++ b/src/jabs/feature_extraction/features.py
@@ -35,7 +35,7 @@ FlatFeatureMap: TypeAlias = dict[str, npt.NDArray[np.generic]]
 PerFrameFeatureMap: TypeAlias = dict[str, dict[str, npt.NDArray[np.generic]]]
 WindowFeatureMap: TypeAlias = dict[str, dict[str, dict[str, npt.NDArray[np.generic]]]]
 
-FEATURE_VERSION = 16
+FEATURE_VERSION = 17
 
 _FEATURE_MODULES = [BaseFeatureGroup, SocialFeatureGroup, SegmentationFeatureGroup]
 
--- a/docs/user-guide/features.md
+++ b/docs/user-guide/features.md
@@ -62,6 +62,10 @@
 * bearing to lixit sine
 * bearing to lixit cosine
 
+When more than one lixit is present, the nearest lixit is selected per frame using the distance
+from the mouse centroid to each lixit tip. Using the centroid rather than a single keypoint (such
+as the nose) keeps this selection stable when individual keypoints are missing from the pose.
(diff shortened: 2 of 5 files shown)


src/jabs/feature_extraction/features.py
```diff
@@ -35,7 +35,7 @@
 PerFrameFeatureMap: TypeAlias = dict[str, dict[str, npt.NDArray[np.generic]]]
 WindowFeatureMap: TypeAlias = dict[str, dict[str, dict[str, npt.NDArray[np.generic]]]]
 
-FEATURE_VERSION = 16
+FEATURE_VERSION = 17
```
keithshep, line 39: Because we're computing a new centroid based lixit feature right?

keithshep (approved):

bergsalex (approved):

src/jabs/feature_extraction/features.py
gbeane: it uses the centroid for determining which limit to use when calculating features, but the features themselves are the same. 

I bumped the version because it could potentially result in different feature values, because it might choose a different lixit as closest lixit due to the nose keypoint dropping out and the nose-lixit distance being corrupted but the centroid-lixit distance would be valid

PR #412, opened 2026-07-14
Defer video/pose frame-count check off project load (KLAUS-505)

gbeane:
## Summary

Removes the up-front video/pose frame-count validation from project load. Today every video file is opened via `cv2.VideoCapture` at load to compare its frame count against the pose file — the dominant, cache-resistant load cost on large projects.

Measured on a 488-video project (external SSD): the video reads were ~24.5s of the load scan (91%), and even a fully warm parallel load went **0.11s → 1.84s** solely because of this check.

## What changed

- `enable_video_check` now defaults to `False`; when `True` it performs the (now opt-in) up-front sweep.
- The check is deferred to when a video is actually opened: `CentralWidget.load_video` shows a non-blocking warning (`MessageDialog`) on a video/pose frame mismatch, at zero extra I/O (the pose and player are already open).
- Feature-extraction workers log a warning on mismatch for headless/batch runs.
- New `get_fps_and_nframes` reads fps + frame count from a single video open.

## Rationale

The training / feature / prediction pipeline is pose-driven — features come from the pose file and labels are indexed to pose frames. A video/pose frame mismatch does not corrupt features or predictions; it only affects frame-display alignment in the labeling GUI. So the correct home for the check is video-open time, not project load.

## Tests

Full suite green. New/updated coverage: the deferred default performs no video opens, the opt-in path (`enable_video_check=True`) still raises on an unreadable video, the worker logs on mismatch, and the GUI mismatch message helper.

Jira: KLAUS-505

### DIFF
--- a/src/jabs/project/parallel_workers.py
+++ b/src/jabs/project/parallel_workers.py
@@ -21,7 +21,7 @@ from jabs.core.constants import MULTICLASS_NONE_BEHAVIOR
 from jabs.core.enums import CacheFormat
 from jabs.pose_estimation import open_pose_file
 from jabs.video_reader import VideoReader
-from jabs.video_reader.utilities import get_fps
+from jabs.video_reader.utilities import get_fps_and_nframes
 
 from .track_labels import TrackLabels
 from .video_labels import VideoLabels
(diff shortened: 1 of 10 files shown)


keithshep (approved):
it all looks good to me

PR #418, opened 2026-07-28
Add JABS Hub-backed projects client integration plan

gbeane:
Adds the client-side design/plan for **JABS Hub-backed projects**: the desktop GUI
(`jabs-behavior-classifier`) becomes a caching client for cloud projects, with video, pose,
annotations, and project metadata stored in JABS Hub and cached locally for frame-accurate labeling,
feature extraction, training, and offline work.

**Design doc only — no code, no dependency changes.** The document is
`docs/development/plans/jabs-hub-backed-projects-plan.md`.

## Companion Hub work

| | |
|---|---|
| Backend spec | `docs/specs/0006-jabs-project-backend-design.md` in `TheJacksonLaboratory/jabs-hub` (merged, #59) |
| Governing ADRs | TheJacksonLaboratory/jabs-hub#61 — ADR-0012 (video identity + artifact addressing), ADR-0014 (project membership authz), ADR-0015 (annotation storage + concurrency). **All `Proposed`, under review.** |

The REST API (spec §7), the data model (spec §5), and the annotation-document schema (Appendix A.1)
are the shared contract between the two repos.

**Two sections track decisions that are not yet ratified**, so expect them to move:

- §2.5's library read/write posture follows ADR-0014 Decision 3
- §5's optional `numFrames` follows ADR-0012 Decision 3

## What reviewers are asked to decide

Eleven decisions are already **Confirmed**. These seven are **Recommended (pending confirm)** — this
is the actual ask:

| # | Decision |
|---|---|
| D2 | MVP conflict policy is optimistic version + last-write-wins, with `VideoLabels.merge` reserved for a later 3-way merge |
| D3 | Annotations stored as a whole document per project-video; normalized per-`(identity, behavior)` is post-MVP *(shared; Hub owns storage)* |
| D4 | `project.json` splits into shared (synced to Hub) vs local/GUI keys |
| D8 | New workspace package `packages/jabs-hub-client` (import `jabs.hub`), depending on `jabs-core` only |
| D22 | Hydration is lazy by default, plus an explicit full-project "make available offline" action (GUI + `jabs-cli`) |
| D23 | Conflict resolution is user-visible, not silent — the losing document is always written to `jabs/hub-conflicts/` |
| D24 | Permission scope: authenticated access + per-project membership enforced server-side, one access level per project for the MVP |

## Where to look

**§8 (Decisions) is the fast path** — it is the whole design in one table. Beyond that:

- **§1** — what a Hub-backed project is, and the phased build order
- **§2.4** — hard requirement that Hub integration stays invisible to non-Hub users: no startup
  network or auth, lazy-imported Hub client, local workflows untouched. The section most likely to
  matter if you do not use Hub.
- **§3** — the existing client seams this builds on (`video_path`, `save_annotations`,
  `ProjectPaths`), with `file:line` references
- **§4** — architecture: media resolver, annotation store, sync engine, local cache
- **§9** — rough effort: ~3.5–4.5 developer-months of client work, 1 developer

## Changed since the 2026-07-30 approval

+165 / −26. @keithshep's approval predates all of it, so a re-review is warranted. Addressed from
that review:

- Markdown and mermaid rendering fixes — the header block relied on trailing-space hard breaks (two
  lines had only one space), and a `;` inside a sequence-diagram message split the statement
- **New §2.5** — permissions and authorization, in scope vs out of scope
- **New conflict-UX spec** in §4.8 (D23)
- **New §2.3 use case 4 + §4.6** — explicit full-project hydration for offline / low-bandwidth work (D22)

Also added since: the library's lab-wide read/write posture plus group-scoped visibility recorded as
a known future requirement (§2.5, §10), and `numFrames` marked optional in the §5 contract with a
probe-on-first-open fallback (§4.6).

### DIFF
--- a/docs/development/plans/jabs-hub-backed-projects-plan.md
+++ b/docs/development/plans/jabs-hub-backed-projects-plan.md
@@ -1,15 +1,15 @@
 # Plan: Hub-Backed JABS Projects — Client Integration (jabs-behavior-classifier)
 
-**Status:** Draft / planning 
-**Author:** Glen Beane (with Claude Code)  
-**Repository:** `jabs-behavior-classifier` (this repo — the PySide6 desktop GUI / Python client).  
-**Companion doc:** the Hub-side (Go) design lives in the `jabs-hub` repo at
-`docs/specs/0006-jabs-project-backend-design.md`. The REST API, the PostgreSQL data model, and
-the annotation-document schema (Appendix A.1) are the **shared contract** — keep them in sync. 
-**Scope:** The client half of Hub-backed JABS projects under a **cloud-native** model: the desktop
-GUI becomes a caching client that pulls videos, pose files, annotations, and project metadata from
-JABS Hub / Google Cloud Storage into a local cache, works against the cache (frame-accurate
-labeling, feature extraction, training, offline), and pushes changes back.
+- **Status:** Draft / planning
+- **Author:** Glen Beane (with Claude Code)
+- **Repository:** `jabs-behavior-classifier` (this repo — the PySide6 desktop GUI / Python client).
+- **Companion doc:** the Hub-side (Go) design lives in the `jabs-hub` repo at
+  `docs/specs/0006-jabs-project-backend-design.md`. The REST API, the PostgreSQL data model, and the
+  annotation-document schema (Appendix A.1) are the **shared contract** — keep them in sync.
+- **Scope:** the client half of Hub-backed JABS projects under a **cloud-native** model: the desktop
+  GUI becomes a caching client that pulls videos, pose files, annotations, and project metadata from
+  JABS Hub / Google Cloud Storage into a local cache, works against the cache (frame-accurate
+  labeling, feature extraction, training, offline), and pushes changes back.
 
 ---
 
@@ -97,6 +97,9 @@ warned.
   expensive one) and a classifier registry for predictions/classifiers are post-MVP (§10).
 - Real-time collaborative cursors / live presence (post-MVP).
 - True frame streaming of remote video (playback is download-to-cache + prefetch; §4.6).
+- **Fine-grained permissions** (roles, per-video/per-behavior rights, group- or lab-level access).
+  The MVP is authenticated access + per-project membership, enforced by Hub — see §2.5 for the full
+  in-scope / out-of-scope breakdown, and D24.
 
 ### 2.3 Motivating use cases
 
@@ -111,6 +114,10 @@ warned.
    project and adds videos to it from the library (JABS recordings or uploaded external files), then
    opens it in the desktop GUI (§4.7) — the media hydrates on demand.
 3. **Work from any machine / HPC** with no manual file copying — the cache hydrates from Hub.
+4. **Going offline, or onto low bandwidth.** Before a flight, a field trip, or a move to a slow or
+   metered link, a user hydrates the **whole** project in one deliberate action — every video, pinned
+   pose file, and annotation document — instead of relying on lazy per-video downloads that would
+   fail once the connection is gone, and pins it so cache eviction leaves it alone (§4.6, D22).
 
 ### 2.4 Unobtrusiveness for non-Hub users (hard requirement)
 
@@ -130,6 +137,50 @@ to them:
 
 A regression test asserts that launching the app and opening a local project make **zero** Hub
 imports or network calls (§7).
+
+### 2.5 Permissions and authorization — what is in and out of scope
+
+Authorization is **enforced by Hub**, not by the client (Hub doc, D5), but the *goals* it has to meet
+belong in scope here so they are not left implicit. Users are **not** assumed to be uniformly
+trustworthy with write access to everything.
+
+**In scope (MVP)**
+
+- **Authenticated access only.** Every Hub request carries an OIDC access token (§4.5); there is no
+  anonymous, shared-secret, or unauthenticated read path. Without a token the client sees no projects.
+- **Per-project membership, enforced server-side on every request.** A project has an owner and a
+  membership set (`project_members`); Hub authorizes each call against the caller's membership. Per
+  D19 projects are **personal (owner-only)** in the first cut, so the MVP membership set is normally
+  one user — but the check is a membership check from day one, so sharing is an added row, not a
+  redesign.
+- **The client is not a trust boundary.** The GUI only *reflects* permissions (e.g. greying out
+  labeling for a project it may not write); it never grants them. A patched client, a hand-edited
+  `hub.json`, or a guessed project ID gains nothing — every read and write is re-authorized by Hub.
+- **Possessing a project identifier is not access.** Pasting a project link (§4.7) only names the
+  project; the caller must still be a member. Links are shareable *because* they confer no rights.
+- **Short-lived, per-request media credentials.** Signed GCS URLs are minted per authorized request
+  and expire; a cached URL is not a durable capability, and copying a cache directory to another
+  machine does not carry Hub access with it.
+- **Attribution comes from the token.** Annotation history records the identity Hub derives from the
+  access token, not the client-supplied `labeler` field (Appendix A.1), which stays a display value.
+
+**Out of scope (MVP) — named so they are deliberate omissions, not oversights**
+
+- **Roles and fine-grained permissions.** No viewer/labeler/owner distinction, no per-behavior or
+  per-video permissions: a project member reads and writes everything in that project. A role model
+  is post-MVP (§10); when it lands, the client's only job is to honor a `permissions` field on the
+  project manifest and disable the corresponding UI.
+- **Lab-, group-, or organization-level access control**, including any mapping from JAX directory
+  groups. Membership is per project and explicit.
+- **Locking as a permission concept.** Concurrent writers are handled by optimistic concurrency
+  (D2/§4.8), not by access control; behavior/video-level locking is post-MVP (§10).
+- **Protecting the local cache.** Cached video, pose, and annotation files are ordinary files under
+  normal filesystem permissions — no client-side encryption at rest, and no attempt to stop a local
+  user from reading a cache they have filesystem access to. Consequence to document for users: do not
+  place a Hub project cache on shared storage whose readers are not all project members. Tokens are
+  the exception — OS keyring, never in project files (§2.4).
+- **Audit and compliance reporting.** Annotation version history provides attribution and rollback,
+  not a tamper-evident audit log.
 
 ---
 
@@ -196,8 +247,8 @@ flowchart LR
         UI[GUI / Project]
         RES[Media resolver]
         AS[Annotation cache/store]
-        CACHE[(Local cache dir\nvideo + pose + jabs/)]
-        DERIVED[(Derived: features/predictions\nrecomputed locally)]
+        CACHE[(Local cache dir<br/>video + pose + jabs/)]
+        DERIVED[(Derived: features/predictions<br/>recomputed locally)]
         HC[HubClient + OIDC]
         UI --> RES
         UI --> AS
@@ -210,8 +261,8 @@ flowchart LR
 
     subgraph Hub["JABS Hub (Go / net-http / api/v0)"]
         API[HTTP API + JWT auth]
-        DB[(PostgreSQL\nvideos, projects, annotations)]
-        GCS[(GCS object storage\nvideo + pose, content-addressed)]
+        DB[(PostgreSQL<br/>videos, projects, annotations)]
+        GCS[(GCS object storage<br/>video + pose, content-addressed)]
         API --> DB
         API --> GCS
     end
@@ -284,8 +335,8 @@ A new workspace package `packages/jabs-hub-client` (namespace `jabs.hub`):
   annotations, media URLs). Maps the `{code,message}` envelope to typed exceptions.
 - **OIDC auth** — Authorization-Code + PKCE against Auth0, token cache + refresh, OS-keyring
   storage. Net-new client capability (the app has no networking today).
-- **Cache manager** — the local cache root, LRU eviction + size cap + "pin for offline," and the
-  sync-state file.
+- **Cache manager** — the local cache root, LRU eviction + size cap + "pin for offline" (D22,
+  including whole-project hydration), and the sync-state file.
 - **Lazy-loaded core dependency.** `jabs.hub` ships as a **core dependency** (always present, so the
   menu action always works) but is **lazy-imported** — it and its networking/auth/keyring
   dependencies load only when a Hub action runs, so non-Hub users pay no import or startup cost
@@ -320,6 +371,21 @@ directory name disambiguates from the Go `jabs-hub` repo while the import stays 
   blake2b hash (a deterministic slice) keys the clip's feature/prediction caches. Annotations are
   clip-relative. *(Client-side offset into a full-source download is the fallback for an
   un-materialized clip.)*
+- **Explicit full hydration — "make available offline" (D22).** Lazy per-video hydration is the
+  default, but the user can also pull an entire project up front for the offline / low-bandwidth case
+  (§2.3 use case 4). **File → "Download Project for Offline Use"** fetches every video, pinned pose,
+  and annotation document in the manifest with progress + cancel, and **pins** the project so the
+  cache LRU (§4.5) will not evict it until it is unpinned. Properties that make this safe to invoke
+  casually:
+  - **Resumable and idempotent** — already-cached, hash-verified files are skipped, so re-running
+    after a cancel, a crash, or a dropped connection fetches only what is missing.
+  - **Budget-aware** — the manifest carries per-video sizes, so the client can state the total up
+    front and warn (with the required vs configured cache cap) before a hydration that would exceed
+    the cache budget, rather than thrashing the LRU mid-download.
+  - **Granular** — the same action exists per video (pin/hydrate a subset) for projects larger than
+    the local cache budget.
+  - **Headless equivalent** — a `jabs-cli` command performs the same hydration for HPC/batch use, so
+    a compute node can pre-warm a cache without launching the GUI.
 - **Library upload is a web-UI / `jabs-cli` task, not a desktop-GUI flow** (§4.7). The upload
   mechanics (compute blake2b, request an upload URL, PUT to GCS, `upload-complete`; pose optional)
   live in `HubClient` and are exercised by a `jabs-cli` helper (to seed/bulk-load the library) and
@@ -377,6 +443,26 @@ See D4.
   the cache so nothing is lost; `VideoLabels.merge` available for a later 3-way merge. Hub enforces
   the check-and-set atomically and returns `409` with the current version (Hub doc —
   Optimistic-Concurrency Contract).
+- **Conflict UX (D23).** Conflicts are rare in the first cut — personal, single-user projects (D19) —
+  but the user-visible behavior is specified now, because "last-write-wins" silently losing a labeling
+  session would be worse than today's manual zip hand-off:
+  - **Never block, never lose.** Labeling never waits on the network; a `409` is resolved on the sync
+    thread. Before applying the resolution the client writes the losing document to
+    `jabs/hub-conflicts/<video>-<version>-<timestamp>.json` in the cache, so a discarded edit is
+    always recoverable from disk (and re-importable via `VideoLabels.load`).
+  - **Notify, don't interrupt.** Resolution surfaces as a non-modal status-bar / sync-indicator
+    message — "Labels for `<video>` were also changed by `<user>`; your version was kept, theirs saved
+    to …" — with a link that reveals the conflict file. No modal dialog appears mid-labeling.
+  - **Sync state is always legible.** A per-project indicator shows `synced` / `syncing` /
+    `offline (N queued)` / `conflict`, with per-video detail on hover, so "did my labels reach Hub?"
+    is answerable at a glance instead of inferred.
+  - **Remote changes don't yank the current video.** For a video the user does not have open, a newer
+    pulled document just updates the cache. For the **currently open** video the client keeps showing
+    the user's labels and offers an explicit **"Reload labels from Hub"** — safe to accept, since the
+    local version is already in the cache and the conflict file.
+  - **Post-MVP:** with `VideoLabels.merge`, disjoint edits (different identity, behavior, or frame
+    range) merge silently and only genuine overlaps surface — the point at which a review dialog
+    showing the conflicting intervals becomes worth building. Deliberately not in the MVP.
 - **Pose consistency is guaranteed** (not just warned): the client pulls the project's **pinned**
   pose file (§4.6), so the labels' identities always align, and the pose changes only via an
   explicit upgrade. The client asserts its cached pose matches the pinned pose file's `poseHash` as a cheap
@@ -399,7 +485,7 @@ sequenceDiagram
     else conflict
         Hub-->>Sync: 409 {currentVersion: M}
         Sync->>Hub: GET annotations (M)
-        Sync->>Sync: resolve (LWW; merge optional)
+        Sync->>Sync: resolve (LWW, merge optional)
         Sync->>Hub: PUT (If-Match: M)
     end
 ```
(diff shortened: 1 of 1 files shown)


docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,608 @@
...
+**Status:** Draft / planning 
+**Author:** Glen Beane (with Claude Code)  
+**Repository:** `jabs-behavior-classifier` (this repo — the PySide6 desktop GUI / Python client).  
+**Companion doc:** the Hub-side (Go) design lives in the `jabs-hub` repo at
+`docs/specs/0006-jabs-project-backend-design.md`. The REST API, the PostgreSQL data model, and
+the annotation-document schema (Appendix A.1) are the **shared contract** — keep them in sync. 
+**Scope:** The client half of Hub-backed JABS projects under a **cloud-native** model: the desktop
+GUI becomes a caching client that pulls videos, pose files, annotations, and project metadata from
+JABS Hub / Google Cloud Storage into a local cache, works against the cache (frame-accurate
+labeling, feature extraction, training, offline), and pushes changes back.
```
keithshep: Some of these are not getting rendered with the newlines implied by formatting (in my vscode MD renderer at least)

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,608 @@
...
+   videos; a colleague trains/evaluates classifiers from those labels. Today the expert zips
+   `<project>/jabs/annotations/` and sends it — manual, overwrites local work, no history, and no
+   guarantee both sides use the *same* pose file. In a cloud project both users open the same Hub
+   project; labels sync automatically with history and attribution, and because the **pose file is
+   pulled from Hub too**, the collaborator's labels always align to the correct identities. No zip,
+   no overwrite, no pose-mismatch footgun.
+2. **Create a project in the web UI, open it in the GUI.** In the JABS Hub web UI a user creates a
+   project and adds videos to it from the library (JABS recordings or uploaded external files), then
+   opens it in the desktop GUI (§4.7) — the media hydrates on demand.
+3. **Work from any machine / HPC** with no manual file copying — the cache hydrates from Hub.
```
keithshep, line 117: will there be a command to hydrate an entire project rather than lazily hydrating videos. The use case I'm thinking of is: I'm going to be offline soon OR I'm going to have lower bandwidth soon

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,608 @@
...
+device/recording/processing platform already exists) **and** an earlier iteration of this plan that
+migrated annotations first and left media local. The cloud-native model was chosen deliberately
+over that piecemeal approach: the video library needs cloud media anyway, and full cloud storage
+makes the label-sharing use case correct (shared pose ⇒ aligned identities) rather than merely
+warned.
+
+---
+
+## 2. Goals, Non-Goals, and Motivating Use Cases
+
```
keithshep, line 71: I think it would be good to be explicit about what kind of permission goals are in scope or out of scope. Eg are we going to assume all users are trustworthy and have write access to all projects or will there be a per-project permission scheme or something else?

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,608 @@
...
+### 4.8 Annotation sync engine + conflict (Phase 3)
+
+- A background **`HubSyncWorker`** (QThread) owns all Hub network I/O; the GUI thread never blocks.
+- On save: write the **annotation cache synchronously** (fast; preserves crash-safety + the
+  training path) → enqueue a **debounced/coalesced** Hub `PUT` with the last-synced `baseVersion`.
+- **Sync-state file** `jabs/hub-sync.json`: per-video `{version, dirty, lastPushedAt}` + an
+  **outbox** for offline replay.
+- On open (online): pull changed annotation docs into the cache (version manifest via
+  `GET /projects/{id}/annotations?includeDocuments=false`). Offline: use the cache.
+- **Conflict:** optimistic version + last-write-wins (matching today's `tmp.replace` semantics) +
```
keithshep, line 506: possibly out of scope for this doc: is it worth specifying user experience in case of conflict?

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,608 @@
...
+  the cache so nothing is lost; `VideoLabels.merge` available for a later 3-way merge. Hub enforces
+  the check-and-set atomically and returns `409` with the current version (Hub doc —
+  Optimistic-Concurrency Contract).
+- **Pose consistency is guaranteed** (not just warned): the client pulls the project's **pinned**
+  pose file (§4.6), so the labels' identities always align, and the pose changes only via an
+  explicit upgrade. The client asserts its cached pose matches the pinned pose file's `poseHash` as a cheap
+  integrity/tamper check.
+
+```mermaid
```
keithshep, line 557: this diagram fails in vscode with:

```
Parse error on line 17:
...LWW; merge optional)        Sync->>Hub:
-----------------------^
Expecting '()', 'SOLID_OPEN_ARROW', 'DOTTED_OPEN_ARROW', 'SOLID_ARROW', 'SOLID_ARROW_TOP', 'SOLID_ARROW_BOTTOM', 'STICK_ARROW_TOP', 'STICK_ARROW_BOTTOM', 'SOLID_ARROW_TOP_DOTTED', 'SOLID_ARROW_BOTTOM_DOTTED', 'STICK_ARROW_TOP_DOTTED', 'STICK_ARROW_BOTTOM_DOTTED', 'SOLID_ARROW_TOP_REVERSE', 'SOLID_ARROW_BOTTOM_REVERSE', 'STICK_ARROW_TOP_REVERSE', 'STICK_ARROW_BOTTOM_REVERSE', 'SOLID_ARROW_TOP_REVERSE_DOTTED', 'SOLID_ARROW_BOTTOM_REVERSE_DOTTED', 'STICK_ARROW_TOP_REVERSE_DOTTED', 'STICK_ARROW_BOTTOM_REVERSE_DOTTED', 'BIDIRECTIONAL_SOLID_ARROW', 'DOTTED_ARROW', 'BIDIRECTIONAL_DOTTED_ARROW', 'SOLID_CROSS', 'DOTTED_CROSS', 'SOLID_POINT', 'DOTTED_POINT', got 'NEWLINE'
```

keithshep (approved):

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a6b6bb0a. The header block was relying on trailing-space hard breaks and two of the five lines only had a single trailing space, so Status+Author and Companion+Scope ran together. Rather than depend on invisible whitespace I converted the whole block to a bullet list, which renders the same everywhere. I also swapped the `\n` line breaks in the §4.1 flowchart node labels for `<br/>` since `\n` in mermaid node text is not documented and renders literally in some versions.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Good call — added. New use case §2.3.4 ("Going offline, or onto low bandwidth") and a spec bullet in §4.6 plus decision D22: lazy per-video hydration stays the default, with an explicit **File → "Download Project for Offline Use"** that pulls every video, pinned pose, and annotation doc in the manifest, with progress + cancel, and **pins** the project so cache LRU eviction leaves it alone. Specified as resumable/idempotent (hash-verified files are skipped, so re-running after a cancel or a dropped connection only fetches what is missing), budget-aware (warns up front if the total exceeds the configured cache cap instead of thrashing the LRU mid-download), granular (per-video pin/hydrate for projects bigger than the cache budget), and with a `jabs-cli` equivalent for headless/HPC pre-warming. Also added a risk row and a test bullet for the resume behavior.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Agreed, that was left implicit. Added §2.5 "Permissions and authorization — what is in and out of scope" (+ D24, a Non-Goals pointer, and a post-MVP entry), and it explicitly does **not** assume all users are trustworthy with write access to everything. In scope for the MVP: authenticated-only access (OIDC token on every request, no anonymous path); per-project membership enforced server-side on every request — per D19 projects are owner-only at first, so the membership set is normally one user, but it is a membership check from day one so sharing is an added row rather than a redesign; the client is explicitly not a trust boundary (it only reflects permissions, never grants them); possessing a project link confers nothing; signed GCS URLs are short-lived per-request, so copying a cache dir does not carry access; attribution comes from the token, not the client-supplied `labeler` field. Out of scope and named as deliberate omissions: roles / fine-grained (per-video, per-behavior) permissions — a member reads and writes everything in the project; group-, lab-, or org-level access control; locking as a permission concept (concurrency is handled by optimistic versioning, D2); protecting the local cache (ordinary files, no encryption at rest — so do not put a project cache on shared storage whose readers are not all members); audit/compliance logging. Authorization enforcement itself stays Hub-owned (Hub doc D5).

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: I think it is in scope — LWW silently eating a labeling session would be worse than the zip hand-off it replaces — so §4.8 now has a "Conflict UX" bullet plus D23. Specified: never block and never lose (labeling never waits on the network; the losing document is always written to `jabs/hub-conflicts/<video>-<version>-<timestamp>.json` in the cache, loadable via `VideoLabels.load`); notify without interrupting (non-modal status-bar/sync-indicator message naming the other user and linking the conflict file — no modal mid-labeling); a persistent sync-state indicator (`synced` / `syncing` / `offline (N queued)` / `conflict`, per-video detail on hover) so "did my labels reach Hub?" is answerable at a glance; remote changes never yank the currently open video — the client keeps showing your labels and offers an explicit "Reload labels from Hub". A real merge-review dialog is deferred to the post-MVP 3-way merge, where only genuine overlaps would surface. Added a test bullet asserting a 409 always leaves a loadable conflict document, and `hub-sync.json` / `hub-conflicts/` rows in the A.2 cache mapping.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a6b6bb0a. Cause was the `;` in `Sync->>Sync: resolve (LWW; merge optional)` — mermaid treats `;` as a statement separator, so it split the line and tried to parse `merge optional)` as a new statement, hence "expecting arrow, got NEWLINE". Changed to a comma; there are no other semicolons inside either mermaid block now. I do not have a mermaid renderer here, so please confirm it parses in your VS Code preview.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+        UI --> RES
+        UI --> AS
+        RES --> CACHE
+        AS --> CACHE
+        CACHE --> DERIVED
+        UI --> HC
+        RES --> HC
+    end
+
+    subgraph Hub["JABS Hub (Go / net-http / api/v0)"]
```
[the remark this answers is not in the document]
gbeane: Fixed in b6ce3090 — `net-http` → `net/http`. No mermaid risk from the slash, since the same label already contains `Go /` and `api/v0`.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+---
+
+## 5. What the client needs from Hub (the contract)
+
+The full API/DTOs/data model are in the Hub doc. The client depends on:
+
+- **Library + media:** `GET /videos` (list/search), `POST /videos` + `.../video-upload-url` /
+  `.../pose-upload-url` / `.../upload-complete` (upload), `GET /videos/{id}/video-url` /
+  `.../pose-url` (signed download). Media is content-addressed; **exact pose bytes preserved**.
+- **Projects & the join manifest:** `GET/POST /projects`, `GET /projects/{id}`; `GET
```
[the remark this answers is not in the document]
gbeane: Fixed in b6ce3090 — now `GET /projects`, `POST /projects`, `GET /projects/{id}`. Agreed it was the odd one out; every other bullet in §5 spells each route separately.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+
+### 3.3 Project + settings seams
+
+- **`Project`** (`project.py:137`) composes `ProjectPaths`, `SettingsManager`, `VideoManager`,
+  `FeatureManager`, `PredictionManager`, `SessionTracker`. **`SettingsManager`**
+  (`settings_manager.py`) owns `project.json` (behaviors, window sizes, settings, metadata,
+  video_files, selected_behavior).
+- Videos are enumerated by a **local directory glob** (`VideoManager.get_videos`,
+  `video_manager.py:153`) → for a cloud project this becomes the Hub **manifest**
+  (`GET /projects/{id}/videos`).
```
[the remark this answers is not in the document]
gbeane: Correct, and fixed in b6ce3090 — `:153` is the `@staticmethod` decorator, `def get_videos` is at `:154`. Updated in both places (§3.3 and Appendix A.3).

Since a pointer had drifted, I audited every `file:line` reference in the doc against the current tree, which turned up one more of the same kind: `video_labels.py:278` is the `@classmethod` line, and `def load` is at `:279`. Fixed in the same commit (three occurrences — §3.2, Appendix A.1, Appendix A.3). Worth noting the Hub-side ADR-0015 already cites `video_labels.py:178,279`, so the two repos now agree.

The other ~25 references check out, including the load-bearing ones: `video_manager.py:251` (`video_path`), `:181` (`get_cached_pose_path`), `:262` (`load_annotations`), `project.py:618` (`save_annotations`), `parallel_workers.py:207` (`_load_video_labels`), `pose_estimation/__init__.py:40` (`get_pose_path`), and `utilities.py:39` (`hash_file`).

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+- **Classifier registry / model cards** — would let the client publish/pull trained classifiers +
+  predictions (the derived artifacts kept local in the MVP, D14).
+- **Cloud feature cache (optional performance tier)** — feature extraction is time-consuming; a
+  future opt-in tier stores pre-computed features in Hub and downloads them instead of recomputing.
+  Safe to share because the cache is keyed by pose hash + `FEATURE_VERSION` + distance scale
+  (§3.1/§4.9), so another machine's features are valid for a byte-identical Hub pose. Client lookup
+  order: local cache → Hub feature cache → compute (optionally upload). Does not change the MVP
+  "recompute locally" default.
+- **Project sharing + role-based permissions** — membership beyond the owner (D19), then a role model
+  (viewer / labeler / owner, potentially per-video or per-behavior) that the client honors via a
```
bergsalex: This is _currently_ named  `VIEWER / EDITOR / ADMIN` in the propsed JABS-hub ADR.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+**In scope (MVP)**
+
+- **Authenticated access only.** Every Hub request carries an OIDC access token (§4.5); there is no
+  anonymous, shared-secret, or unauthenticated read path. Without a token the client sees no projects.
+- **Per-project membership, enforced server-side on every request.** A project has an owner and a
+  membership set (`project_members`); Hub authorizes each call against the caller's membership. Per
+  D19 projects are **personal (owner-only)** in the first cut, so the MVP membership set is normally
+  one user — but the check is a membership check from day one, so sharing is an added row, not a
+  redesign.
+- **Membership gates projects, not the video library.** A library video exists before and
```
bergsalex: As currently written, ADR-0014 Decision 6 makes library access the union of video visibility and membership in any containing study or project. So project membership now does grant library access to that project's videos.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+  independently of any project (ADR-0012), so project membership is not a well-defined question for
+  library media: **ADR-0014 Decision 3 makes the library readable *and writable* by any authenticated
```
bergsalex: Reads still effectively open under the LAB default and creating a video is unaffected, but writes to an existing video now need owner or study/project membership.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+- **Roles and fine-grained permissions.** No viewer/labeler/owner distinction, no per-behavior or
+  per-video permissions: a project member reads and writes everything in that project. A role model
+  is post-MVP (§10); when it lands, the client's only job is to honor a `permissions` field on the
+  project manifest and disable the corresponding UI.
+- **Lab-, group-, or organization-level access control**, including any mapping from JAX directory
+  groups. Membership is per project and explicit.
+- **Group-scoped library visibility — out of scope for the MVP, but a known future requirement, not a
+  hypothetical.** Library videos will eventually need to be readable only to specified groups
+  (collaborator data, embargoed studies), so the lab-wide catalogue above is a starting posture rather
+  than the end state. Enforcement is Hub-side (ADR-0014 Decision 3 today, its OQ2 for the ACL model);
```
bergsalex: The ACL model now has a concrete shape `(visibility, owner_subject)`.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+  "recompute locally" default.
+- **Project sharing + role-based permissions** — membership beyond the owner (D19), then a role model
+  (viewer / labeler / owner, potentially per-video or per-behavior) that the client honors via a
+  `permissions` field on the project manifest; group- and lab-level access control. Explicitly out of
+  scope for the MVP (§2.5, D24).
+- **Group-scoped library visibility (a committed requirement, not a maybe).** Library videos readable
+  only to specified groups — see §2.5. Client-side work when it lands: honor a visibility field on the
+  library listing and the project manifest, present "no longer accessible" states for cached media
+  whose access was withdrawn, and keep every library browse path behind a filtered query rather than a
+  full scan. Sequenced after Hub decides where group membership lives (ADR-0014 OQ1/OQ2), which is the
```
bergsalex: OQ2 is now OQ3 and OQ1 is resolved.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+
+---
+
+## 5. What the client needs from Hub (the contract)
+
+The full API/DTOs/data model are in the Hub doc. The client depends on:
+
+- **Library + media:** `GET /videos` (list/search), `POST /videos` + `.../video-upload-url` /
+  `.../pose-upload-url` / `.../upload-complete` (upload), `GET /videos/{id}/video-url` /
+  `.../pose-url` (signed download). Media is content-addressed; **exact pose bytes preserved**.
```
bergsalex: This is true only for the upload path.

A pipeline's output prefix is write-once and keyed by run_id on purpose (ADR-0010 DAG-8), so a re-run never overwrites a prior result.

docs/development/plans/jabs-hub-backed-projects-plan.md
```diff
@@ -0,0 +1,747 @@
...
+  Optimistic-Concurrency Contract).
+- **Conflict UX (D23).** Conflicts are rare in the first cut — personal, single-user projects (D19) —
+  but the user-visible behavior is specified now, because "last-write-wins" silently losing a labeling
+  session would be worse than today's manual zip hand-off:
+  - **Never block, never lose.** Labeling never waits on the network; a `409` is resolved on the sync
+    thread. Before applying the resolution the client writes the losing document to
+    `jabs/hub-conflicts/<video>-<version>-<timestamp>.json` in the cache, so a discarded edit is
+    always recoverable from disk (and re-importable via `VideoLabels.load`).
+  - **Notify, don't interrupt.** Resolution surfaces as a non-modal status-bar / sync-indicator
+    message — "Labels for `<video>` were also changed by `<user>`; your version was kept, theirs saved
```
bergsalex: A Hub change since this was written adds a second reason for 409.

If someone unlinks and relinks a video in the web UI the old association's annotations are dropped and a new, empty one replaces it (spec 0006 §7.3). A client holding queued offline edits for that video then pushes against an association that has no annotations. It follows the new association because sync state is keyed by video name and resolved to a `projectVideoId` from the manifest.

Hub used to accept that push silently as version 1. It now returns 409.

The client already writes the rejected edit to hub-conflicts/ and reports it without interrupting the user, so the handling itself needs no change. Only the message needs a second case.

Perhaps:
> "Labels for `<video>` no longer exist on Hub (the video was re-linked); your version was saved to …"

bergsalex (approved):

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a. §2.5 and §10 now name `VIEWER` / `EDITOR` / `ADMIN` (ADR-0014 Decision 2), and §10 records the `OWNER` rename with the reason — the lab means *point of contact* by the word, which Decision 3 splits out as attribution carrying no access — plus a note not to surface "owner" as a permission level.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a. §2.5 replaced the old "membership gates projects, not the library" framing with Decision 6's union predicate, and states your consequence as its own bullet: **project membership *does* grant read access to that project's videos**, through the union's `project_videos ⋈ project_members` clause — not as a side effect of an open catalogue.

One qualification added later in cd1eba85: the predicate is specified but not built. On `jabs-hub@adr-0012-implementation` there is no `visibility` / `owner_subject` / `project_members` at all, and `GET /videos` is authenticated-only, so §2.5 now reads as the target rather than current behavior, with the delta tracked in the new §5.1.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a. §2.5 carries both halves: reads effectively lab-wide under the `LAB` default, and a separate bullet that **writes to an existing library video are narrower** — its `owner_subject`, or a study/project member with write access — with creating a video unaffected and `POST /videos/{id}/pose-runs` gated at write level (Decision 7).

cd1eba85 adds that none of it is built yet (§5.1), so reads are currently open for a stronger reason than the permissive default, and notes that the write rules do not constrain the desktop client, which neither creates library videos nor starts pose runs.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a, refined in cd1eba85. §10 now says ADR-0014 Decision 6 specifies `visibility` + `owner_subject` on `videos` plus the union predicate, so the ACL model is "no longer a blank" — replacing the old text that called it an open question. The refinement was tense: the columns are specified, not shipped, so §10 points at §5.1 for what exists today.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a. §10 now cites **ADR-0014 OQ3** for *where group membership lives*, and drops the OQ1/OQ2 reference. Verified against `jabs-hub@adr-0012-implementation`: OQ3 is the group-membership question (OQ1 is the lab-level-admin question, OQ2 is what bounds GPU spend), and resolving it adds `'GROUP'` to the `visibility` CHECK plus one predicate clause — which §10 now states.

§10 also records the knock-on effect you flagged elsewhere: sharing is no longer gated on the same decision, because subject discovery is resolved by the `users` registry and `GET /users?q=`. The two were coupled only while that was open.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a. §5 now scopes it: content-addressed storage (`cas/{blake2b}`) covers the **upload path only**, and a pipeline-produced pose is catalogued **in place** at its write-once, run-keyed derived prefix (`{pipeline}/{version}/{run_id}/…`, ADR-0010 DAG-8, so a re-run never overwrites a prior result) and is never copied into `cas/`.

It also states why this costs the client nothing: identity comes from the manifest's `contentHash`/`poseHash`, the client follows signed URLs, and it never constructs a storage key — which matches migration `0016`'s own consumer contract.

docs/development/plans/jabs-hub-backed-projects-plan.md
gbeane: Fixed in a1ca5d1a, using your suggested wording. §4.8 now documents both `409` causes, discriminated by `details.currentVersion`: a concurrent write (`M`), and a vanished lineage (`null`) after an unlink/relink, citing spec 0006 §7.3 for the cascade and ADR-0015's principle that a write carrying a base version must never win against a state it was not derived from. The sequence diagram gained the second branch, and D23 was rewritten to cover both.

Two things beyond the message, both following your point that the mechanics need no second code path:

- The client deliberately does **not** auto-repush the queued edit as a fresh `version 1` — a relink may have re-pinned a different pose file or clip range, so silently restoring labels keyed to the old pose would reintroduce the identity-misalignment footgun the plan exists to remove. Recovery is an explicit action from the conflict file.
- Verified the discriminator against ADR-0015's own upsert: `current_version` is `(SELECT version FROM cur)`, so it is null exactly when no annotation row exists, while a first-write race yields `1`.

PR #424, opened 2026-08-05
Add feature cache status scanning with train/classify warnings

gbeane:
## Summary

Adds an in-memory picture of a project's on-disk feature cache and uses it in two places: the video **Get Info** dialog, and a warning before **Train** / **Classify** when features are not cached for the selected window size (they would otherwise be computed mid-run, which can take minutes per video with no explanation to the user).

## What's new

**Cache inspection (`jabs-io`)** — `inspect_identity_cache()` reports what one per-identity cache directory holds: format, feature version, pose hash, cached window sizes, whether per-frame features are present, size on disk. It reads only metadata (HDF5 attributes or `metadata.json`), never feature data, and never validates: an unreadable cache is reported the same as an absent one. Parquet window sizes are only reported when the `window_{size}.parquet` file actually exists, so the result reflects what could really be loaded.

**Per-video status (`jabs.project.feature_cache_status`)** — `VideoFeatureCacheStatus` aggregates the per-identity summaries: `window_sizes` (cached for every cached identity), `partial_window_sizes`, `cached_identity_count` / `is_complete`, `cache_formats`, `is_stale`, `cm_units`, `size_bytes`, and `has_window_features(window_size, identities=None)`. Both on-disk layouts are handled: flat `features/<video>/<id>/` and the CLI's `features/<video>/<hash>/<id>/`.

Nothing here validates a cache against its pose file: that would mean hashing every pose file, which is far too slow for a status scan. A cache whose pose file changed still reports as present and is detected when the features are actually loaded.

**Background scan at project open** — `FeatureCacheScanThread` scans the project off the main thread when it opens (the scan touches every per-identity cache directory, slow enough on network storage to matter) and the results are stored on the `Project`. The thread supports cooperative termination and is stopped in `MainWindow.closeEvent`.

**Project API** — `feature_cache_status`, `set_feature_cache_status()`, `refresh_feature_cache_status(video)`, `invalidate_feature_cache_status()`, `videos_missing_window_features(...)`, and `labeled_identities(behaviors)`. Videos with no stored status are scanned on demand, so the queries never depend on the background scan having finished. Training and classification invalidate the status when they finish, since a run caches whatever it computed.

**Get Info** — new **Feature Cache** section: cache directory, identities cached, window sizes (and partial ones), format, feature version with an out-of-date flag, distance units, size on disk, and a warning row when per-frame features are missing. The dialog re-scans the single video so what it shows is current.

**Train / Classify warning** — a Yes/No dialog naming the number of videos whose features are missing, with the video list in the details section. The scope differs per action so the warning is accurate rather than noisy: training checks only *labeled* identities (it never reads features for unlabeled ones), while classification checks every identity of every target video.

## Verification

- 960 tests pass (878 before, 82 added), plus 237 in `jabs-io`. Ruff clean.
- Checked against real `IdentityFeatures` output for both cache formats: the scanner agrees with what the production write path produces (3 identities, window sizes 5 and 10, correct version, not stale).
- Checked the warning scope end to end on a real project: with nothing cached, Train flags the labeled video and Classify flags both; after a training run caches the labeled identity, Train no longer warns while Classify still does (it needs the other four identities); a different window size warns again.
- Drove a real offscreen `MainWindow` through the background scan to confirm results cross the thread boundary and land on the project.

## Docs

`gui.md` (both the online copy and the in-app copy) gets a **Feature Computation** section and an expanded **Get Info** description.

### DIFF
--- a/containers/docker/Dockerfile.pose
+++ b/containers/docker/Dockerfile.pose
@@ -25,10 +25,11 @@ COPY packages/jabs-core ./packages/jabs-core
 COPY packages/jabs-io ./packages/jabs-io
 COPY packages/jabs-vision ./packages/jabs-vision
 
-# Install jabs-vision (with the HRNet extra) and its workspace deps, incl. h5py.
+# Install jabs-vision (with the HRNet extra) and its workspace deps. h5py comes in
+# as a required dependency of jabs-io.
 RUN uv pip install --system \
       "./packages/jabs-core" \
-      "./packages/jabs-io[h5py]" \
+      "./packages/jabs-io" \
       "./packages/jabs-vision[hrnet_msfork]"
 
 ENTRYPOINT ["jabs-pose"]
--- a/docs/user-guide/gui.md
+++ b/docs/user-guide/gui.md
@@ -26,7 +26,7 @@ Right-click a video name to open a context menu:
 
 <img src="imgs/video-context-menu.png" alt="Video list context menu" width=300 />
 
-- **Get Info:** Open a dialog showing details about the video and its associated pose file.
+- **Get Info:** Open a dialog showing details about the video, its associated pose file, and its cached features (the cache directory, cached window sizes, storage format, size on disk, and whether the cache is out of date).
 - **Copy Video Name:** Copy the video's filename to the clipboard.
 - **Classify Video:** Run the trained classifier for the current behavior on just this video, rather than every video the way the **Classify** button does. This item is only available once a classifier has been trained for the current behavior (if none is ready, JABS prompts you to train one first). When classification finishes, JABS displays the new predictions, switching to the video if it is not already the active one.
 - **Exclude from Training:** Toggle whether the video is held out of classifier training (unchecked by default). When a video is excluded:
@@ -48,6 +48,12 @@ Right-click a video name to open a context menu:
 - **Symmetric Behavior Toggle:** Tells the classifier that the behavior is symmetric. A symmetric behavior is when left and right features are interchangeable.
 - **All k-fold Toggle:** Uses the maximum number of cross validation folds. Useful when you wish to compare classifier performance and may have an outlier that can be held-out.
 - **Cross Validation Slider:** Number of "Leave One Out" cross validation iterations to run while training.
+
+### Feature Computation
+
+Training and classification both need extracted features for the selected window size. When a project is opened, JABS scans the project's feature cache in the background to find out which videos already have them. If features are missing when you click **Train** or **Classify**, JABS warns you first: the run will still work, but the features are computed as it goes, which can take several minutes per video. You can continue anyway or cancel and precompute features for the whole project with the `jabs-features` command line tool.
+
+Training needs features only for the subjects you have labeled, while classification needs them for every subject in every video it processes, so **Classify** may warn when **Train** does not. Computed features are cached, so the warning goes away once a run has computed them. Use **Get Info** in the video list's right-click menu to see exactly what is cached for a video.
 
 ### Choosing a Classifier Type
 
--- a/docs/user-guide/gui.md
+++ b/docs/user-guide/gui.md
@@ -51,7 +51,7 @@ Right-click a video name to open a context menu:
 
 ### Feature Computation
 
-Training and classification both need extracted features for the selected window size. When a project is opened, JABS scans the project's feature cache in the background to find out which videos already have them. If features are missing when you click **Train** or **Classify**, JABS warns you first: the run will still work, but the features are computed as it goes, which can take several minutes per video. You can continue anyway or cancel and precompute features for the whole project with the `jabs-features` command line tool.
+Training and classification both need extracted features for the selected window size. When a project is opened, JABS scans the project's feature cache in the background to find out which videos already have them. If features are missing when you click **Train** or **Classify**, JABS warns you first: the run will still work, but the features are computed as it goes, which can take several minutes per video. You can continue anyway, or cancel and precompute features for the whole project with the `jabs-init` command line tool (`jabs-init -w <window size> <project directory>`).
 
 Training needs features only for the subjects you have labeled, while classification needs them for every subject in every video it processes, so **Classify** may warn when **Train** does not. Computed features are cached, so the warning goes away once a run has computed them. Use **Get Info** in the video list's right-click menu to see exactly what is cached for a video.
 
--- a/packages/jabs-io/README.md
+++ b/packages/jabs-io/README.md
@@ -28,8 +28,8 @@ save(data_instance, 'frames.parquet')
 | `nwb`     | `pynwb`, `ndx-pose` | NWB pose file read/write  |
 | `parquet` | `pyarrow`           | Parquet feature cache I/O |
 
-`jabs-io` depends on `jabs-core`, which requires `h5py` directly, so HDF5 support is
-always available with no extra needed. `pyarrow` is also a direct dependency of
+HDF5 is not an extra: `h5py` is a required dependency, since pose files, prediction
+files and the default feature cache are all HDF5. `pyarrow` is a direct dependency of
 `jabs-behavior-classifier`, so the `parquet` backend is available automatically when
 `jabs-io` is installed as part of the full JABS application. The `nwb` extra must
 always be installed explicitly:
--- a/packages/jabs-io/pyproject.toml
+++ b/packages/jabs-io/pyproject.toml
@@ -15,6 +15,10 @@ authors = [
 dependencies = [
     "jabs-core",
     "numpy>=2.0.0,<3.0.0",
+    # HDF5 is fundamental to JABS: pose files, prediction files and the default
+    # feature cache are all HDF5, and most modules in this package import h5py
+    # unconditionally. Declared directly rather than relied on transitively.
+    "h5py>=3.10.0,<4.0.0",
 ]
 
 [tool.uv.sources]
@@ -48,9 +52,6 @@ nwb = [
     "ndx-pose>=0.2.2",
     "ndx-multisubjects>=0.1.1",
     "pynwb>=3.1.3",
-]
-h5py = [
-    "h5py>=3.15.1",
 ]
 parquet = [
     "pyarrow>=18.0.0",
--- a/packages/jabs-io/src/jabs/io/base.py
+++ b/packages/jabs-io/src/jabs/io/base.py
@@ -16,10 +16,7 @@ except ImportError:
     pa = None
     pq = None
 
-try:
-    import h5py
-except ImportError:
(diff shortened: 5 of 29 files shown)


gbeane:
Follow-up commit (d77999a): the in-memory status was being invalidated wholesale on run completion and rebuilt lazily, which meant the repopulating scan ran synchronously on the GUI thread inside the next Train/Classify click handler — a visible stall on a large project on network storage.

Two changes:

- **Invalidation is now scoped to what the run touched.** Training records the labeled videos it will read (`_training_cache_targets`, set once the warning is accepted) and invalidates only those; classification invalidates `_classification_targets` (`None` still means every video). The rest of the map stays warm.
- **Repopulation happens off the main thread.** `CentralWidget.feature_cache_changed` is emitted after invalidation and `MainWindow` connects it to the existing `refresh_feature_cache_status()`, so the background scan rebuilds the status while the user reads the training report. If they click Train before it finishes, the on-demand rescan in `videos_missing_window_features()` still covers it.

Because scan results replace the whole map, `_feature_cache_scan_complete()` now ignores results from a superseded scan (an earlier scan still finishing would otherwise write back what the cache looked like before the run computed features).

Still no eager update at the moment features are written: the workers that write them don't report back what they wrote, and the rescan is metadata-only.

Verified on a real project with real caches on disk, driving a real offscreen MainWindow: after a single-video classification finishes, only that video's status is dropped, the background rescan repopulates both videos, the newly computed features are reflected (`videos_missing_window_features(5) == []`), and a superseded scan's results are ignored. 969 tests pass (9 added), ruff clean.

gbeane:
all three were real. Addressed in cd42238.

**1. Concurrent scan threads (`refresh_feature_cache_status`).** Correct, and the shutdown consequence was the serious part: `closeEvent` only terminates the tracked thread, so an untracked earlier scan could still be running at teardown — the exact `QThread: Destroyed while thread is still running` abort that #425 exists to prevent. It was reachable in practice now that every finished training/classification run triggers a refresh.

Fixed by coalescing to a single in-flight scan: when a refresh arrives while one is running, that scan is asked to stop (so its now-stale results are never emitted) and a replacement is queued, started from `_feature_cache_scan_finished`. `closeEvent` clears the queue before waiting, so shutdown can't kick off a new scan.

Measured with real threads, five refreshes fired back to back and each video's scan slowed down:

| | max concurrent scans | scan passes |
| --- | --- | --- |
| before | 5 | 5 |
| after | 1 | 2 (the interrupted one + the coalesced rescan) |

**2. `is_complete` and per-frame features across pose-hash directories.** Also correct, and it was inconsistent with `window_sizes`, which already merges per identity. Added `identities_missing_per_frame`, which folds per-frame presence per identity (an identity is satisfied if any of its cache directories has per-frame features); `is_complete` now uses it. The Get Info warning row used the same per-directory `any(...)` test and had the same flaw, so it now uses the property too — and names the affected identities instead of saying "one or more".

**3. `cache_formats` sort key.** Right: `str(CacheFormat.HDF5)` is `'CacheFormat.HDF5'`, not `'hdf5'`, so the code didn't match the docstring. Now sorts on `.value`. Same order today, but it no longer depends on `Enum.__str__`.

Tests added: per-identity per-frame merging (including the multi-pose-hash case), `cache_formats` ordering by value, the scan queue/coalescing behavior and its shutdown interaction, `_start_feature_cache_scan` thread setup, and the dialog's warning row (present and absent). 977 tests pass, ruff clean.

gbeane:
Addressed in 43e0c70 (plus the two suppressed comments). Details, including where I landed differently than suggested:

**Shutdown wait ignoring `wait()`'s return value.** Real gap — confirmed by simulating storage that doesn't respond (scan thread blocked in an `open()` syscall): quitting logged `QThread: Destroyed while thread is still running` and the process exited **134 (SIGABRT)**.

I did not adopt the "longer/blocking wait" part of the suggestion: an unbounded wait trades a harmless exit-time abort for an unquittable app, on exactly the slow/unresponsive storage this feature targets. The cooperative stop is checked between videos and cannot interrupt a filesystem call that never returns, so waiting longer only postpones the same outcome.

Instead `closeEvent` now delegates to `_stop_feature_cache_scan()`, which escalates: ask the scan to stop → `wait(2000)` → if that returns `False`, log a warning and `terminate()` → `wait(1000)` → if even that fails, log an error. Terminating is safe here in a way it usually isn't: the scan only reads cache metadata, so it cannot corrupt project data, and losing the scan costs nothing but the status it would have produced.

Two alternatives I tested and rejected:
- Detaching the thread from the window (`setParent(None)` + a module-level reference) so its destructor wouldn't run: **still aborts** — Python takes ownership and deletes it during finalization.
- An unbounded `wait()`: hangs the quit indefinitely.

Measured on the wedged-storage simulation, quitting via the window close button:

| | close() duration | exit code |
| --- | --- | --- |
| before | 2.0 s | 134 (SIGABRT, `QThread: Destroyed while thread is still running`) |
| after | 2.0 s | 0, with `Feature cache scan did not stop within 2000 ms (unresponsive storage?); terminating the thread` |

A responsive scan is unaffected: it stops cooperatively in ~0.1 s and is never terminated.

**Suppressed comment: `labeled_identities()` reads every annotation file on the Train click.** Fair. The check now short-circuits: `_confirm_training_features()` first asks the cheap whole-video question (in-memory status, no annotation I/O), and only when something is missing does it read annotations to narrow to labeled identities. Since the labeled identities are a subset of all identities, an empty whole-video result means the narrowed result is empty too, so this never changes the answer — it just makes the fully-cached case (a project prepared with `jabs-features`, the recommended workflow) free. When features genuinely are missing we do read the annotation files, which is the only way to know whether the warning is real; that cost is small next to the run the user is about to start, which re-reads all of them anyway.

**Suppressed comment: `_cleanup_classify_thread()` docstring said "training thread".** Pre-existing copy-paste; corrected to "classification thread".

985 tests pass, ruff clean. New tests cover the escalation ladder (responsive stop, forced stop, unstoppable thread logged) and the short-circuit (asserting `labeled_identities` is not called when everything is cached).

gbeane:
Addressed in 4ae32e2, but by going the other way on the premise: **h5py is now a required dependency of `jabs-io`, not an optional extra.**

HDF5 isn't a format-specific backend in JABS — pose files, prediction files and the default feature cache are all HDF5 — and the "optional" extra was already fiction. Five modules in this package import h5py unconditionally at module scope, and have for a long time:

- `io/feature_cache/hdf5.py`
- `io/internal/pose/hdf5.py`
- `io/internal/prediction/hdf5.py`
- `io/internal/dataclass/hdf5.py`
- `io/feature_cache/inspection.py` (this PR)

Only `io/base.py` guarded it, and its guard was unreachable: every `HDF5Adapter` subclass lives in one of those modules, so you cannot import a subclass without h5py to hit the "install the h5py extra" error. The package README already stated HDF5 needs no extra (h5py arrived transitively via `jabs-core`, which requires it) — the extra just made the dependency look optional while every HDF5 code path assumed it.

So rather than adding a lazy import and `h5py is None` handling to a module that would still be dead code:

- `h5py>=3.10.0,<4.0.0` moved into `jabs-io`'s `dependencies`, matching the bound `jabs-core` uses; the `h5py` optional-dependency group is removed.
- `io/base.py` imports h5py unconditionally like its siblings, and the unreachable `HDF5Adapter.__init__` guard is gone.
- `containers/docker/Dockerfile.pose` no longer installs `./packages/jabs-io[h5py]` (that extra no longer exists; h5py comes in as a required dependency).
- `uv.lock` regenerated: `jabs-io` now lists `h5py` directly. Verified by resolving `jabs-io` on its own into a clean venv — `h5py==3.16.0` is pulled in.

**Suppressed comment: `_directory_size()` could raise from `iterdir()`.** Valid, and fixed — I had guarded `stat()` but not the listing, so an unreadable or vanishing cache directory could have escaped the "unreadable cache behaves like absent" contract and aborted a project-wide scan. The listing is now guarded too and contributes a size of 0; a test asserts inspection still returns its other fields when the listing is denied.

**Suppressed comment: handle `h5py is None` in `_inspect_hdf5_cache()`.** Not applicable after the above — there is no `None` case, since h5py is a hard requirement and imported unconditionally.

983 tests on this branch plus 238 in `jabs-io`, ruff clean.

gbeane:
Both fixed in 6ea9a7b.

**`has_window_features()` treating unusable caches as coverage.** Correct, and the miss is bigger than it first looks: bumping `FEATURE_VERSION` is routine in this project (contributors are told to bump it whenever features change), so after any release that does, every cache on disk is stale — and the gate would have stayed silent through the full recompute it exists to announce. Same for an identity whose per-frame features are missing.

The fix separates the two questions the status object answers:

- **What is on disk** — `window_sizes`, `partial_window_sizes`, `is_stale`, `is_complete` are unchanged. Get Info should keep showing "window sizes: 5, 10" next to "Feature version: 16 (out of date, current is 17)"; reporting "none" there would hide data that really is present.
- **What would load** — `has_window_features()` now goes through `_loadable_window_sizes_by_identity()`, which only counts a cache written by the current feature version and holding per-frame features. `Project.videos_missing_window_features()` inherits this, so the warning fires.

One subtlety worth calling out: the loadable sizes are merged across an identity's pose-hash subdirectories *per directory*, not per field. A window size present only in a stale directory no longer counts because a sibling directory happens to be current — no single cache there could serve the run.

Verified against real caches built by `IdentityFeatures` (Parquet, the default for new projects): with current caches `videos_missing_window_features(5) == []`; after aging the recorded version by one, the same status still reports `window_sizes == (5,)` and `is_stale`, and the video is reported as needing computation.

Two causes of recomputation still can't be seen from a status scan, and I've documented them on the method rather than leaving them implicit: a pose file that changed since the cache was written (detecting it means hashing every pose file, too slow for a status scan) and a distance-unit setting that no longer matches the cache. The second one is detectable — `cm_units` is already on the status object, it just needs the run's setting passed in — so if you want that gap closed too, say so and I'll add it.

**`_cleanup_classify_thread()` leaving `_classification_targets` set.** Right, and the asymmetry was mine: `_cleanup_training_thread()` consumes its target list but this one didn't. The completion path happened to clear it afterwards, but the cancel and failure paths (`_classify_thread_error_callback`) did not, so a canceled run left the stale list on the widget. The cleanup now owns that lifecycle and clears it, and the redundant assignment in `_classify_thread_complete()` is gone (it reads the targets into a local first, before the cleanup).

While adding tests I also fixed a latent trap in the ones I wrote earlier: `test_project_feature_cache_status.py` hard-coded feature version 17 while the code compares against the real `FEATURE_VERSION`, so the next bump would have broken those tests for no good reason. The helper now defaults to `FEATURE_VERSION` and takes an explicit older version when a test wants a stale cache.

989 tests on this branch, ruff clean.

PR #427, opened 2026-08-07
Fix IndexError in delete_bouts and suppress a lone too-short bout in the duration filter

gbeane:
## Reasoning

I looked for candidates across all six categories and picked this one because it is a category-1 bug (unhandled edge case) with a reproducible crash, it sits in non-GUI core logic (`jabs-behavior`), and the fix is small and testable. Other candidates I considered and passed over, all lower priority:

- `docs/development/development.md` omits `jabs-behavior` from the "Project Structure" tree and the "Monorepo Transition" list (category 2, docs-only).
- The three postprocessing stages' `help()` docstrings say `Returns: FilterHelp` but the actual type is `StageHelp` (category 2, cosmetic).
- `GapInterpolationStage.__init__` reassigns `self._config = {...}` where the sibling stages assign a key into the dict inherited from `super().__init__` (category 3, currently equivalent).

### What was wrong

[`BehaviorEvents.delete_bouts`](https://github.com/KumarLabJax/JABS-behavior-classifier/blob/main/packages/jabs-behavior/src/jabs/behavior/events/behavior_events.py#L57-L120) assumed every bout marked for removal had a neighbor to merge into. When the run-length encoding held a single bout, the `cur_gap == 0` branch did `new_durations[1] += new_durations[0]` on a length-1 array and raised `IndexError`.

All three postprocessing stages reach this when an identity's prediction vector is a **single run that is also below the stage's threshold**. Both conditions are required — a uniform vector alone is not enough, since a whole-video run of thousands of frames never satisfies `durations < min_duration`:

```python
# crashes on main
BoutDurationFilterStage(min_duration=5).apply(np.array([1, 1, 1]), None)
GapInterpolationStage(max_interpolation_gap=5).apply(np.array([-1, -1, -1]), None)
BoutStitchingStage(max_stitch_gap=3).apply(np.array([0, 0]), None)
# IndexError: index 1 is out of bounds for axis 0 with size 1

# does NOT crash — uniform, but the run exceeds the threshold
BoutDurationFilterStage(min_duration=5).apply(np.full(1800, 1), None)
```

So the real triggers are a very short clip, or a threshold configured larger than the video — a latent edge case rather than something users are likely hitting today. The same path is also reached when *every* bout is marked for removal, though that is unreachable from the stages (RLE runs alternate by construction, so a stage filtering on one class can never mark them all).

Neither call site guards against it: `BinaryClassifyStrategy.postprocess_identity` (`src/jabs/ui/classify_strategy.py:132`) and the per-identity loop in `src/jabs/scripts/cli/postprocessing.py:388` both let the `IndexError` propagate uncaught.

### Why the fix is safe

The `delete_bouts` guard only fires when fewer than two bouts remain — exactly the states that previously raised. Every input that worked before takes an identical path.

Leaving the lone bout in place *inside the primitive* is the length-preserving choice: `delete_bouts` absorbs a removed bout's frames into its neighbors, so with no neighbor there is nothing to absorb them, and deleting anyway would return a vector shorter than the input. Callers index `postprocessed[i]` into a fixed `(n_identities, n_frames)` array and require the length to be preserved.

But the primitive is state-agnostic, so it cannot decide the *outcome*, and the three stages do not want the same one. `BoutDurationFilterStage` exists to eliminate brief predictions, so a rejected bout surviving is the filter failing at its job — and with only the guard, the same 3-frame bout got opposite treatment depending on whether anything else was in the vector:

```
[1,1,1,0,0,0] -> [0 0 0 0 0 0]   suppressed
[1,1,1]       -> [1 1 1]         kept
```

Interpolation and stitching *do* want the guard's behavior: an all-NONE vector has no prediction to spread into the gap, and a vector with no BEHAVIOR bouts has no gaps to join. So the outcome decision belongs in the stages, which is what the second commit does.

### Logic changes

- **`packages/jabs-behavior/src/jabs/behavior/events/behavior_events.py`** — added a `len(new_durations) < 2` guard at the top of the deletion loop in `delete_bouts`, which skips bouts that have no neighbor left to merge into. Also corrected two errors in the same docstring:
  - it documented a `Returns:` value, but the method mutates in place and returns `None`;
  - its note on odd-length interpolation said the *previous* state gets the extra frame. The code assigns `floor(n/2)` to the previous bout and `ceil(n/2)` to the next, so the *next* state gets it. The existing test `test_delete_bouts_non_matching_borders_odd_duration` already asserts the code's actual behavior, so only the prose was wrong.

- **`packages/jabs-behavior/src/jabs/behavior/postprocessing/stages/duration_stage.py`** — `apply()` returns `np.full_like(classes, NOT_BEHAVIOR)` when the whole vector is one bout it marked for removal, matching what a neighboring NOT_BEHAVIOR bout would have done. `full_like` preserves the caller's dtype and length.

- **Tests** (`test_behavior_events.py`, `test_duration_filter.py`, `test_interpolation_stage.py`, `test_stitching_stage.py`) — regression coverage for: deleting the only bout; marking every bout for removal; the with-neighbor/without-neighbor pair asserting they now agree; a bout exactly at `min_duration` (the comparison is strict `<`); a lone short NONE bout the duration filter ignores; and the lone bout being *kept* by interpolation and stitching, so the difference between the three stages reads as a decision rather than an oversight.

### Mechanical updates

None. No imports, exports, or public signatures changed, and no `FEATURE_VERSION` bump (no computed feature values are affected).

### Verification

`ruff check .`, `ruff format .`, `pytest packages/jabs-behavior/tests` (109 passed) and the root suite (806 passed, 200 skipped) all clean.

---

This PR was produced by an automated analysis from Claude Code.

### DIFF
--- a/packages/jabs-behavior/src/jabs/behavior/events/behavior_events.py
+++ b/packages/jabs-behavior/src/jabs/behavior/events/behavior_events.py
@@ -57,14 +57,19 @@ class BehaviorEvents:
     def delete_bouts(self, indices_to_remove) -> None:
         """Helper function to delete events from bout data.
 
+        The deleted bout's frames are absorbed by its neighboring bouts, so the
+        total number of encoded frames is preserved. This object is modified in
+        place.
+
         Args:
             indices_to_remove: event indices to delete
 
-        Returns:
-            Bouts object that has been modified to interpolate within deleted events
-
         Notes:
-            Interpolation on an odd number will result with the "previous" state getting 1 more frame compared to "next" state
+            Interpolation on an odd number of frames will result with the "next" state getting 1
+            more frame compared to the "previous" state.
+
+            A bout with no neighbor to absorb its frames (i.e. the only remaining bout) is left
+            in place, since removing it would shorten the encoded vector.
         """
         new_durations = np.copy(self.durations)
         new_starts = np.copy(self.starts)
@@ -72,7 +77,11 @@ class BehaviorEvents:
         if len(indices_to_remove) > 0:
             # Delete backwards so that we don't need to shift indices
             for cur_gap in np.sort(indices_to_remove)[::-1]:
-                if cur_gap == 0:
+                if len(new_durations) < 2:
+                    # Nothing left to merge this bout into. Deleting it anyway would drop its
+                    # frames from the encoded vector, so leave it as-is.
+                    continue
+                elif cur_gap == 0:
                     # Remove the first bout, merge with the next (use next bout's class)
                     new_durations[1] += new_durations[0]
                     new_starts[1] = new_starts[0]
--- a/packages/jabs-behavior/src/jabs/behavior/postprocessing/stages/duration_stage.py
+++ b/packages/jabs-behavior/src/jabs/behavior/postprocessing/stages/duration_stage.py
@@ -25,6 +25,14 @@ class BoutDurationFilterStage(PostprocessingStage):
     def apply(self, classes: np.ndarray, probabilities: np.ndarray) -> np.ndarray:
         """Apply the duration filter to the predictions.
 
+        A too-short bout is normally removed by absorbing its frames into the
+        neighboring bouts, which for a lone short BEHAVIOR bout means the
+        surrounding NOT_BEHAVIOR state takes them over. When the whole vector is
+        one too-short BEHAVIOR bout there is no neighbor to absorb it, so the
+        frames are set to NOT_BEHAVIOR directly: leaving them as BEHAVIOR would
+        keep a bout this filter rejected, and would make the result depend on
+        whether other frames happened to be present.
+
         Args:
             classes (np.ndarray): The predicted classes.
             probabilities (np.ndarray): The predicted probabilities. (Not used in this stage.)
@@ -39,9 +47,13 @@ class BoutDurationFilterStage(PostprocessingStage):
             rle_data.states == ClassLabels.BEHAVIOR,
         )
 
-        if np.any(bouts_to_remove):
-            rle_data.delete_bouts(np.where(bouts_to_remove)[0])
+        if not np.any(bouts_to_remove):
+            return rle_data.to_vector()
 
+        if len(rle_data.states) == 1:
+            return np.full_like(classes, ClassLabels.NOT_BEHAVIOR)
+
+        rle_data.delete_bouts(np.where(bouts_to_remove)[0])
         return rle_data.to_vector()
 
     @classmethod
--- a/packages/jabs-behavior/tests/test_behavior_events.py
+++ b/packages/jabs-behavior/tests/test_behavior_events.py
@@ -232,6 +232,35 @@ class TestBehaviorEvents:
         np.testing.assert_array_equal(events.durations, [2, 2, 2])
         np.testing.assert_array_equal(events.states, [0, 1, 0])
 
+    def test_delete_bouts_only_bout_is_kept(self):
+        """Test delete_bouts on a single-bout encoding leaves the bout in place.
+
+        There is no neighboring bout to absorb the frames, so removing it would shorten the
+        encoded vector.
+        """
+        events = BehaviorEvents.from_vector(np.array([1, 1, 1]))
+        events.delete_bouts([0])
+
+        np.testing.assert_array_equal(events.starts, [0])
+        np.testing.assert_array_equal(events.durations, [3])
+        np.testing.assert_array_equal(events.states, [1])
+        np.testing.assert_array_equal(events.to_vector(), [1, 1, 1])
+
+    def test_delete_bouts_all_bouts_keeps_last_remaining(self):
+        """Test delete_bouts when every bout is marked for removal.
+
+        The bouts merge until a single bout is left, which is then kept so the encoded vector
+        keeps its original length.
+        """
+        events = BehaviorEvents.from_vector(np.array([0, 0, 1, 1]))
+        events.delete_bouts([0, 1])
+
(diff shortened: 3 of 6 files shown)


gbeane:
Pushed e5e6fd0 on top of this. The `delete_bouts` guard is the right crash fix and I left it alone; what it can't do is decide the *outcome*, and for one of the three stages the outcome it produced was wrong.

## The inconsistency

With just the guard, the same 3-frame short BEHAVIOR bout got opposite treatment depending on whether anything else was in the vector:

```
[1,1,1,0,0,0] -> [0 0 0 0 0 0]   short behavior suppressed
[1,1,1]       -> [1 1 1]         short behavior kept
```

`BoutDurationFilterStage` exists to "eliminate spurious or brief predictions", so a bout below `min_duration` surviving is the filter failing at its job — and the frames that survive are the positive class, the direction that matters downstream.

The reason the guard can't fix this is layering: `delete_bouts` is a generic RLE primitive with no idea what the states mean, so "leave the lone bout in place" is the only choice available to it. That decision belongs to the caller, and the three callers do not agree on it:

- **duration filter** — a lone too-short BEHAVIOR bout should become NOT_BEHAVIOR, which is exactly what a neighboring NOT_BEHAVIOR bout would have done to it.
- **gap interpolation** — an all-NONE vector has no prediction to spread into the gap, so the frames stay NONE. The guard's behavior is already right here.
- **bout stitching** — no BEHAVIOR bouts means no gaps to join, so the frames stay NOT_BEHAVIOR. Also already right.

Two stages want the guard's behavior and one wants the opposite, which is the tell that the branch belongs in the stages.

## What I changed

- `BoutDurationFilterStage.apply()` returns `np.full_like(classes, NOT_BEHAVIOR)` when the whole vector is one bout it marked for removal. Post-processing is binary-only (`MultiClassClassifyStrategy.postprocess_identity` returns `None`), so NOT_BEHAVIOR is unambiguous, and `full_like` preserves the caller's dtype and length.
- Rewrote the stage regression test for the corrected expectation, and added: the with-neighbor/without-neighbor pair asserting they now agree, a bout exactly at `min_duration` (the comparison is strict `<`), and a lone short NONE bout (not a BEHAVIOR bout, so untouched).
- Added one test each to the interpolation and stitching stages asserting the lone bout *is* kept, with the reason. Without those, the difference between the three stages looks like an oversight instead of a decision.
- Annotated `test_delete_bouts_all_bouts_keeps_last_remaining` to say the surviving state follows from merge order rather than a rule. It is order-dependent — `[0,0,1,1]` gives all 0s and `[1,1,0,0]` gives all 1s — and it is unreachable from any stage, since RLE runs alternate by construction so a stage filtering one class can never mark every bout. Worth keeping as an API test, but not worth treating as a contract.

`ruff check`, `ruff format`, `pytest packages/jabs-behavior/tests` (109 passed) and the root suite (1006 passed) are clean.

## One thing to fix in the description

> All three postprocessing stages hit this whenever an identity's prediction vector is uniform.

Uniform is not sufficient — the run also has to be shorter than the stage's threshold. A whole-video vector for an identity is thousands of frames, so `durations < min_duration` is false and nothing is marked:

```
duration filter on 1800 all-BEHAVIOR frames: unchanged   # no crash
```

The real triggers are a very short clip or a threshold configured larger than the video. That makes this a latent edge case rather than something users are hitting, which is worth stating accurately given the description leads with "reproducible crash".

gbeane:
Checked e5e6fd0 out and verified each claim rather than taking it on faith. All of it holds.

**The inconsistency was real.** Reproduced on 39d94be, and the pair now agrees on e5e6fd0:

```
lone short BEHAVIOR    : [0 0 0]
short BEHAVIOR+neighbor: [0 0 0 0 0 0]
```

**The layering argument is right, and it's the part I got wrong.** I treated "don't shorten the vector" as the whole answer, but that's only the primitive's constraint — it says nothing about which state should own the frames. `delete_bouts` can't answer that because it never sees what the states mean, so the guard silently picked "keep" for all three callers when only two of them wanted it. Putting the decision in the stage is the correct split, and the two tests asserting interpolation and stitching *do* keep the bout are what make that legible; without them the asymmetry looks accidental.

Also confirmed: strict `<` means a lone bout at exactly `min_duration` is kept; the lone-short-NONE case is untouched by the duration filter; and the merge-order dependence is exactly as described — `[0,0,1,1]` → all 0s, `[1,1,0,0]` → all 1s. Agreed it isn't worth treating as a contract given it's unreachable from the stages.

**You're right about the description.** "Whenever an identity's prediction vector is uniform" was too strong — the run also has to be under the threshold, and 1800 uniform BEHAVIOR frames pass through untouched. I've rewritten that section to state both conditions, show the non-crashing counterexample, and call it a latent edge case rather than leading with "reproducible crash." Title and change list updated for the second commit too.

One observation, pre-existing and not a blocker. The justification for NOT_BEHAVIOR cites `MultiClassClassifyStrategy.postprocess_identity` returning `None`, which covers the GUI path — but `jabs-cli postprocessing` applies stages straight to `pred.predicted_class` without checking `pred.class_names`, so a multi-class prediction file has no such gate there. That doesn't weaken the change: the `states == ClassLabels.BEHAVIOR` mask directly above already assumes binary semantics, so `full_like(NOT_BEHAVIOR)` adds no new assumption. It's a gap in the CLI's input validation, worth its own issue if you want it tracked.

Re-ran on e5e6fd0: `ruff check` clean, `pytest packages/jabs-behavior/tests` 109 passed, root suite 806 passed / 200 skipped. (My 200 skips are the display-dependent GUI tests — that's the difference from your 1006.)

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #449, opened 2026-09-01
Export video with pose overlay from the GUI and jabs-cli

gbeane:
Adds **File → Export Video with Pose Overlay…** and **`jabs-cli export-video`**, both writing a copy of a video with the pose overlay burned into every frame.

Manually tested end to end in the GUI, plus a full render through the CLI.

## ⚠️ Unrelated change included: segmentation contour width

`src/jabs/video_reader/frame_annotation.py` narrows the segmentation contour from 2px to 1px:

```diff
-    cv2.drawContours(img, trimmed_contours, -1, color, 2)
+    cv2.drawContours(img, trimmed_contours, -1, color, 1)
```

This is a deliberate visual preference, and it is **not** part of the purpose of this PR. Flagging it explicitly because it is easy to miss in a diff this size, and because it is not export-only: `draw_all_contours` is reached through `overlay_segmentation`, whose callers are

- `player_widget/player_thread.py:132` - the live player overlay (View → Overlay Segmentation)
- `video_export/frame_renderer.py:65` - the export path added here

So **every user sees thinner contours in the player**, not just in exported videos. If you would rather this landed on its own, say so and it can be split into a separate one-line PR.

## Behaviour

Pose skeleton is always drawn; segmentation contours are optional (a checkbox in the GUI, `--segmentation/--no-segmentation` on the CLI), defaulting to on so the output matches what **Export Frame** already produces. Unchecking gives a pose-only video. The checkbox is disabled with an explanatory tooltip for pose files older than v6, which carry no segmentation data.

Every identity is drawn identically at full opacity — an export has no active identity, so no animal is singled out. Whole video only; no frame-range option.

The CLI works on any video/pose pair and needs no JABS project; it finds the highest-version pose file beside the video unless `--pose-file` says otherwise.

CLI example:
image

## One renderer, not two

`jabs/video_export/frame_renderer.py` is now the single compositing path behind Export Frame, Export Video and the CLI — `PlayerWidget.get_overlay_frame()` was refactored to call it, so a still and a frame of an exported video come from the same code. A separate cv2 reimplementation of the skeleton would have drifted, and an export that doesn't match the player is worse than no export.

A useful consequence: painting targets a `QImage` rather than a `QPixmap`, and `QImage` needs no `QGuiApplication`. The CLI therefore needs no Qt bootstrapping, no display and no offscreen platform plugin.

## Module moves

Two moves were required, not cosmetic:

**`pose_drawing` out of `jabs.ui`.** `jabs/ui/__init__.py` imports `MainWindow`, so anything importing from under `jabs.ui` drags in the whole GUI — `video_export` → `jabs.ui.player_widget.pose_drawing` → back into `player_widget` failed outright. It is now a top-level package (`skeleton.py` + `colors.py`), which also let `KEYPOINT_COLOR_MAP` live in its own module instead of being folded into the drawing code. Its `__init__` docstring records why each alternative home was rejected, including `jabs.video_reader` — that package is imported by `jabs.project.parallel_workers`, which runs in process-pool workers, so putting Qt behind it would add Qt's import cost to every worker spawn. Verified `from jabs.video_reader import VideoReader` still loads no Qt.

The keypoint palette is byte-identical to before the move; a silent colour shift is not something a test would have caught, so it was checked explicitly.

**`video_export` as a package**, splitting the frame renderer from the video writer, with tests mirroring the split.

## Progress dialog

The first GUI build left the progress dialog stuck on screen — `CustomProgressDialog.closeEvent()` ignores close events by design, which also makes `close()` a silent no-op for the owner.

That dialog exists to be undismissable, which is right for training and classification because they mutate project state. Cancelling an export just stops writing one file, so this uses a plain `QProgressDialog` instead — it auto-closes at 100% and treats the window close button as cancel, making the failure mode structurally impossible rather than merely avoided. That also matches `check_for_updates` in the same file.

`tests/ui/test_progress_dialog.py` pins `CustomProgressDialog`'s contract so the trap is discoverable and the no-dismiss guarantee is protected, and `create_cancelable_progress_dialog`'s docstring now says when to reach for it.

## Notes

- Default codec is `mp4v`. `avc1` (H.264) gives smaller files but is frequently absent from `opencv-python-headless` on Linux, where it fails at writer-open time; it is available via `--codec avc1`.
- Cancelling deletes the partial output rather than leaving an unplayable file.
- The progress bar's maximum comes from the container's frame count. If that over-reports, the bar stops short of 100% and auto-close doesn't fire, but the completion handler closes the dialog explicitly.

## Testing

1097 root tests (+31), all four package suites pass, `ruff check` and `ruff format --check` clean.

Beyond the unit tests: the CLI was run against a generated v3 pose file and a rendered frame inspected visually (two identities, twelve distinctly-coloured keypoints, white skeleton). Both the completion and cancel paths were driven with a real `VideoExportThread` against a real dialog to confirm the dialog is dismissed and the partial file removed.

Docs updated in both user-guide trees, verified still byte-identical.


src/jabs/scripts/cli/export_video.py
```diff
@@ -0,0 +1,140 @@
...
+        raise click.ClickException(f"{output_path} already exists. Use --force to overwrite.")
+
+    if pose_path is None:
+        try:
+            pose_path = get_pose_path(video_path)
+        except ValueError as e:
+            raise click.ClickException(str(e)) from e
+
+    console = Console()
+    pose_est = open_pose_file(pose_path)
```
bergsalex: Might make sense to wrap this in a try/except to give the user a more helpful error.

src/jabs/ui/main_window/menu_handlers.py
```diff
@@ -184,6 +189,128 @@ def export_frame(self) -> None:
...
+
+        if dialog.exec() != QtWidgets.QFileDialog.DialogCode.Accepted:
+            return  # user cancelled
+        selected_files = dialog.selectedFiles()
+        if not selected_files:
+            return
+
+        output_path = Path(selected_files[0])
+        if output_path.suffix.lower() != ".mp4":
+            output_path = output_path.with_suffix(".mp4")
```
bergsalex: According to Claude, filenames with a dot in them will mutate here.
e.g. `session_2024.09.01` turns into `session_2024.09.mp4`

src/jabs/video_reader/frame_annotation.py
```diff
@@ -169,7 +169,7 @@ def draw_all_contours(img: np.ndarray, seg_data: np.ndarray, color: tuple[int, i
         None
     """
     trimmed_contours = trim_seg_list(seg_data)
-    cv2.drawContours(img, trimmed_contours, -1, color, 2)
+    cv2.drawContours(img, trimmed_contours, -1, color, 1)
```
bergsalex, line 172: Claude wants me to mention that this is not _just_ export related.

src/jabs/video_export/video_writer.py
```diff
@@ -0,0 +1,152 @@
...
+                "%d frame(s) will be written without an overlay",
+                video_path.name,
+                total_frames,
+                pose_est.num_frames,
+                max(0, total_frames - pose_est.num_frames),
+            )
+
+        output_path.parent.mkdir(parents=True, exist_ok=True)
+        writer = cv2.VideoWriter(
+            str(output_path), cv2.VideoWriter_fourcc(*codec), fps, (width, height)
```
bergsalex, line 154: > A --codec value that isn't exactly 4 chars (mp4, h2641) blows up in VideoWriter_fourcc with a TypeError before the isOpened() check can produce the documented VideoExportError. A len(codec) != 4 guard right above this would keep the contract.

bergsalex (approved):

src/jabs/scripts/cli/export_video.py
gbeane: Good call - done. `open_pose_file()` is now wrapped, so a malformed pose file gives `Could not read pose file <path>: <reason>` instead of an h5py/OSError traceback. Covered by `test_unreadable_pose_file_is_a_clean_error`.

src/jabs/ui/main_window/menu_handlers.py
gbeane: Confirmed and fixed - `Path('session_2024.09.01').with_suffix('.mp4')` really does yield `session_2024.09.mp4`. Now appends instead of replacing, so the name the user typed is never rewritten. Two tests added: the dotted case, and that an existing `.mp4` isn't doubled up.

src/jabs/video_export/video_writer.py
gbeane: Confirmed - `VideoWriter_fourcc(*'mp4')` raises `TypeError: missing required argument 'c4'` before `isOpened()` is reached. Added a `len(codec) != 4` guard raising `VideoExportError`, so the documented contract holds. Parametrised test covers too-short, too-long and empty.

src/jabs/video_reader/frame_annotation.py
gbeane: yeah, I thought it looked better with a smaller width so I snuck this in 🤣

gbeane: Answered - you were right, and it is the second reading. The 2px→1px contour change is a deliberate visual preference that is not part of this PR's purpose, and `draw_all_contours` is reached from `player_thread.py:132` as well as the new export path, so the live player overlay changes for everyone too.

Keeping it here by choice, but it is no longer buried: the PR description now carries a section calling it out, naming both call sites and the fact that it is not export-only. Happy to split it into its own one-line PR instead if you would prefer to judge the visual on its own.

PR #462, opened 2026-09-09
Fix nwbinspector findings in NWB pose export

ptuan5:
Static-object PoseEstimationSeries (corners, lixit, etc.) stored a single timestamp with 2 coordinates, shape (1, 2), which nwbinspector's check_data_orientation always flags as CRITICAL since the non-time axis is longer than the time axis. Store the constant value at two timestamps spanning the session (start and end) instead, so the shape becomes (2, 2) while the value stays unchanged; the reader only uses data[0] so round-tripping is unaffected.

Also add descriptions to every PoseEstimationSeries/TimeSeries (keypoints, static/dynamic objects, identity mask, bounding boxes) to clear the check_description "placeholder" suggestions, and add support for a keywords session-metadata field (NWBFile accepts it, but nothing in this pipeline could set it before).


packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -343,6 +352,11 @@ def _write_multisubject(self, data: PoseData, path: Path, **kwargs) -> None:
                         name=_bounding_box_key(name),
                         data=data.bounding_boxes[i],  # (num_frames, 2, 2)
                         unit="pixels",
+                        description=(
+                            f"Per-frame bounding box for identity '{name}': "
+                            "[[upper_left_x, upper_left_y], [lower_right_x, lower_right_y]]; "
+                            "-1 indicates a missing value"
```
gbeane: **`-1 indicates a missing value` is wrong — missing bounding boxes are `NaN`.**

`PoseEstimationV8` builds the per-identity bbox array with `np.full((num_frames, num_identities, 2, 2), np.nan, dtype=bboxes.dtype)` (`src/jabs/pose_estimation/pose_est_v8.py:73-74`), and the writer here passes `data.bounding_boxes[i]` through unmodified. So frames with no detection are `NaN` in the NWB file, not `-1`.

`-1` is the **Parquet** convention, not the NWB one — see `src/jabs/scripts/cli/convert_parquet.py:97`: `# Replace NaN with -1 as placeholder for missing bounding box coordinates`. It looks like the sentinel was copied from there.

Concrete failure: export a v8 pose file where identity 1 is absent for frames 0-99. Those rows are `NaN`, but the file's own description tells a downstream consumer to look for `-1`. A consumer following the description treats `NaN` as a valid box (and would treat a genuine `-1` from a `convert_parquet`-derived pose file as missing). Because this string is written into every exported file it cannot be corrected after the fact.

Same text at line 454 in `_write_per_identity`. Suggest either describing the actual sentinel (`NaN indicates no detection for this identity in this frame`) or normalizing `NaN` -> `-1` on write if `-1` is the intended wire format.

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
...
+                session-end timestamp.
+            num_frames: Total number of frames in the session, used to compute
+                the session-end timestamp.
         """
+        if fps <= 0:
+            raise ValueError(f"fps must be positive, got {fps}")
+        if num_frames <= 0:
+            raise ValueError(f"num_frames must be positive, got {num_frames}")
+        end_time = num_frames / fps
+        timestamps = [0.0, end_time]
```
gbeane, line 1015: **The format documentation still describes the single-timestamp layout this replaces — including a section telling users the warning is expected.**

Five documents still specify the old layout and are now wrong:

- `docs/user-guide/nwb-export.md:329-346` — *the file this PR edits*. It states each static object has a "**single timestamp (`t = 0.0 s`)**", that each series "has data shape `(1, 2)`", that "the DANDI validator will emit a `NWBI.check_data_orientation` warning for each static keypoint", and then: "**These warnings are expected and can be ignored.** ... it fires a false positive for static objects, which legitimately have only one timestamp by definition." That paragraph now describes a problem this PR removed and should be deleted.
- `src/jabs/resources/docs/user_guide/nwb-export.md:329-346` — the in-app GUI copy, same text.
- `docs/user-guide/file-formats.md:261-262` and `src/jabs/resources/docs/user_guide/file-formats.md:261` — "**single timestamp (`t = 0.0 s`)**".
- `docs/development/jabs-nwb-format.md:194-205` — the format spec's field table, which still gives `data` as shape `(1, 2)` and `timestamps` as `[0.0]`.

Cost: the spec no longer matches what the writer produces, and users are still told to ignore a warning that no longer fires — which was the whole point of the change.

docs/user-guide/nwb-export.md
```diff
@@ -131,6 +132,7 @@ files and otherwise defaults to the time the export was run.
 | `institution`            | string                    | Institution name.                                                                                                                                                                         |
 | `experiment_description` | string                    | Free-text description of the experiment.                                                                                                                                                  |
 | `session_id`             | string                    | Lab-specific session identifier.                                                                                                                                                          |
+| `keywords`               | list of strings           | Keywords/tags describing the session.                                                                                                                                                     |
```
gbeane, line 135: **The in-app copy of this document was not updated.**

The user-guide docs exist in two places that must be kept in sync: `docs/user-guide/` (online docs) and `src/jabs/resources/docs/user_guide/` (in-app GUI help). This PR adds the `keywords` example and this table row to `docs/user-guide/nwb-export.md` only; `src/jabs/resources/docs/user_guide/nwb-export.md` (lines 121 and 132 hold the identical example block and table) is untouched.

Cost: users reading the help browser inside JABS get no mention of `keywords` and cannot discover the new field.

src/jabs/scripts/cli/convert_to_nwb.py
```diff
@@ -181,6 +181,7 @@ def pose_to_pose_data(
         "institution",
         "experiment_description",
         "session_id",
+        "keywords",
```
gbeane, line 184: **`keywords` is forwarded to pynwb with no type validation, unlike its only other non-string sibling.**

The forwarding loop below (`for key in _SESSION_METADATA_FIELDS - {"session_start_time"}`) copies the raw JSON value straight into `write_kwargs`, and `_make_nwb_file` passes it to `NWBFile` unchecked. pynwb declares it as `{'name': 'keywords', 'type': 'array_data', ...}` (`pynwb/file.py:352`), and hdmf's `array_data` macro does not include `str`.

Concrete failure: a session metadata file containing `{"keywords": "open field"}` — a natural mistake, since every other documented flat key except `session_start_time` *is* a plain string — loads and validates fine, the entire pose HDF5 is read and converted, and only then does hdmf raise an opaque docval `TypeError`, surfaced by `cli.py` as `ClickException(str(e))`. `session_start_time`, the only other field that is not a plain string, gets an explicit `_parse_session_start_time` validator with a friendly message; `keywords` should get the same (reject non-list, and non-string elements).

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
...
+            fps: Frames per second of the source video, used to compute the
+                session-end timestamp.
+            num_frames: Total number of frames in the session, used to compute
+                the session-end timestamp.
         """
+        if fps <= 0:
+            raise ValueError(f"fps must be positive, got {fps}")
+        if num_frames <= 0:
+            raise ValueError(f"num_frames must be positive, got {num_frames}")
+        end_time = num_frames / fps
```
gbeane: **Off-by-one: `num_frames / fps` is one frame period *past* the end of the session.**

Every other series in the file is written with `rate=float(fps)` and the default `starting_time=0`, so frame `k` sits at `k / fps` and the last sample (frame `num_frames - 1`) sits at `(num_frames - 1) / fps`. This end timestamp is `num_frames / fps`, one full frame period beyond it.

Concrete failure: 300 frames at 30 fps. Keypoint series span `[0.0, 9.9667]`; the static object's second sample lands at `10.0`. A consumer mapping the static object back to a frame index computes `10.0 * 30 = 300`, which is out of range for a 300-frame video. The docstring above claims "t=session end" and "both ends of the session", which is not what this computes.

`(num_frames - 1) / fps` makes the static object end exactly where every other series ends.

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
...
         Args:
             name: Name for this PoseEstimation (matches the static object key).
             points: Shape (N, 2) array of x, y coordinates.
             skeleton: Skeleton with N nodes, one per point.
+            fps: Frames per second of the source video, used to compute the
+                session-end timestamp.
+            num_frames: Total number of frames in the session, used to compute
+                the session-end timestamp.
         """
+        if fps <= 0:
```
gbeane, line 1007: **This validation is in the wrong place: it only runs when static objects exist, and leaves the real `rate=0` problem unfixed.**

`PoseData.__post_init__` (`packages/jabs-core/src/jabs/core/types/pose.py`) validates array shapes but never checks `fps > 0` or `num_frames > 0`. Putting the check here — in a static-object-only helper, reached only because this is the one code path that happens to divide by `fps` — produces two inconsistent behaviors for the same bad input:

- `PoseData(fps=0, static_objects={"corners": ...})` -> the whole export hard-fails with `ValueError: fps must be positive`.
- `PoseData(fps=0, static_objects={})` -> exports "successfully", writing `rate=0.0` on every `PoseEstimationSeries` and `TimeSeries` in the file. That is exactly what nwbinspector's `check_rate_is_not_zero` / `check_rate_is_positive` flag, so the file this PR is trying to make clean still fails validation, silently.

The deeper fix is to validate `fps > 0` (and a non-zero frame count) once in `PoseData.__post_init__`, or at the top of `write()`, so every series benefits and the behavior does not depend on whether the pose file happens to contain a lixit. As written the check is also unreachable from the CLI, since `open_pose_file` never passes `fps` and `PoseEstimation.__init__` defaults it to 30.

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -1040,12 +1088,16 @@ def _build_dynamic_object_pose_estimation(
             slot_confidence = (dyn_obj.counts > slot).astype(np.float64)
             for kp in range(n_keypoints):
                 series_name = f"{name}_{slot}" if n_keypoints == 1 else f"{name}_{slot}_{kp}"
+                description = f"Dynamic object '{name}' detection slot {slot}"
```
gbeane, line 1094: **Dynamic objects hit the identical `check_data_orientation` failure and were not fixed.**

`data=dyn_obj.points[:, slot, kp, :]` a few lines below has shape `(n_predictions, 2)`. When `n_predictions == 1` that is `(1, 2)` — the exact shape this PR went to some trouble to eliminate for static objects — and `check_data_orientation` fires CRITICAL again.

This is reachable: fecal boli predictions are made on a sparse subset of frames, so a short clip (or a coarse sampling interval) yields a single sample index. The same `(1, 2)` / `(1, 2, 2)` shape also occurs for keypoint series and bounding-box `TimeSeries` on a one-frame pose file.

Since the root cause is shared, consider handling it once (e.g. a small helper that pads a length-1 sample axis, or at least noting in the PR why dynamic objects are out of scope) rather than special-casing the static-object path only.

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -733,6 +760,7 @@ def _make_nwb_file(**kwargs) -> NWBFile:
             "institution",
             "experiment_description",
             "session_id",
+            "keywords",
```
gbeane, line 763: **No tests were added for any of the new behavior.**

The diff touches no test file. Specifically untested:

- `keywords` reaching the NWB file. `packages/jabs-io/tests/internal/pose/test_nwb.py:894` (`test_session_metadata_fields_written_to_nwb`) asserts every *other* session field round-trips — `lab`, `institution`, `experimenter`, `experiment_description`, `session_id` — and was not extended with `keywords`. Adding `keywords=["open field"]` there plus `assert list(nwb.keywords) == ["open field"]` is a two-line change.
- The new static-object shape and timestamps. Existing tests (`test_static_objects_roundtrip`, `test_static_objects_nwb_structure`, `test_static_objects_multiple_points`) only check round-tripped values and container names, so they pass either way and would not catch a regression back to `(1, 2)` — i.e. nothing pins the behavior this PR exists to establish.
- The two new `ValueError` paths (`fps <= 0`, `num_frames <= 0`).

CLAUDE.md: "Every new module must include tests. PRs adding untested core code won't pass review." and "Focus on core business logic: feature extraction, classifiers, pose processing, **I/O adapters**."

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
         name: str,
         points: npt.NDArray,
         skeleton: Skeleton,  # type: ignore[valid-type]
+        fps: float,
+        num_frames: int,
     ) -> PoseEstimation:
-        """Build a single-timestamp PoseEstimation for a static spatial object.
+        """Build a PoseEstimation for a static (unchanging) spatial object.
```
gbeane, line 987: **Docstring is missing the `Raises:` section for the two new `ValueError`s.**

This method now raises `ValueError` in two places (lines 1007-1010) but the docstring documents only `Args:`. CLAUDE.md, Docstrings: "Include `Args`, `Returns`, and `Raises` sections as applicable; omit empty sections."

The new failure mode is also absent from the two public entry points that can surface it: `write()`'s `Raises:` block (lines 186-191) lists only the identity-name collision `ValueError` and the `ImportError`, and `run_conversion`'s `Raises:` in `src/jabs/scripts/cli/convert_to_nwb.py` lists only the pose-file and `session_start_time` errors. A caller reading the public docstring has no way to know an export can now fail on `fps`.

(While here: the method returns a `PoseEstimation` but has no `Returns:` section either — its sibling builders `_build_pose_estimation` and `_build_dynamic_object_pose_estimation` both have one.)

gbeane (commented):
Code review of the nwbinspector fixes. The `(1, 2)` -> `(2, 2)` change does correctly clear `check_data_orientation` (it tests `any(data_shape[1:] > data_shape[0])`, so `(2, 2)` passes), and 2 timestamps stays under `check_regular_timestamps`'s `len(timestamps) > 2` guard, so no new best-practice warning is introduced. 12 findings below, the first three being the most important: one factually incorrect description that gets permanently embedded in every exported file, and the format documentation still describing the layout this PR just replaced.

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
...
+            raise ValueError(f"num_frames must be positive, got {num_frames}")
+        end_time = num_frames / fps
+        timestamps = [0.0, end_time]
         series_list = [
             PoseEstimationSeries(
                 name=f"{name}_{i}",
-                data=points[i : i + 1, :].astype(np.float64),  # shape (1, 2)
-                confidence=np.ones(1, dtype=np.float64),
+                data=np.tile(points[i].astype(np.float64), (2, 1)),  # shape (2, 2)
+                confidence=np.ones(2, dtype=np.float64),
```
gbeane, line 1020: **The repeat count `2` is hardcoded in three places that must stay in sync.**

`timestamps = [0.0, end_time]` (2 entries), `np.tile(..., (2, 1))`, and `np.ones(2, ...)`. Changing the number of sample points — e.g. if a future nwbinspector check wants 3+ timestamps — requires finding all three, and a mismatch between `data`'s first axis and `confidence`/`timestamps` produces a `check_timestamps_match_first_dimension` failure or an hdmf shape error rather than anything obvious at this call site.

Deriving it once is equivalent and self-consistent:

```python
timestamps = [0.0, end_time]
n_samples = len(timestamps)
...
data=np.tile(points[i].astype(np.float64), (n_samples, 1)),
confidence=np.ones(n_samples, dtype=np.float64),
```

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
         name: str,
         points: npt.NDArray,
         skeleton: Skeleton,  # type: ignore[valid-type]
+        fps: float,
```
gbeane, line 984: **`fps: float` diverges from the `int` used everywhere else for this value.**

`PoseData.fps` is declared `fps: int`, and both sibling builders annotate it as `int` — `_build_pose_estimation(..., fps: int, ...)` (line ~917) and `_build_dynamic_object_pose_estimation(..., fps: int, ...)` (line ~1065). Both call sites pass `fps=data.fps`, an `int`.

Not a runtime problem (PEP 484's numeric tower accepts `int` where `float` is annotated), but it makes the helper look like it accepts fractional frame rates that the rest of the type surface rules out, and it is the one place a reader has to stop and check. `fps: int` for consistency.

packages/jabs-io/src/jabs/io/internal/pose/nwb.py
```diff
@@ -952,24 +981,43 @@ def _build_static_object_pose_estimation(
...
+        if num_frames <= 0:
+            raise ValueError(f"num_frames must be positive, got {num_frames}")
+        end_time = num_frames / fps
+        timestamps = [0.0, end_time]
         series_list = [
             PoseEstimationSeries(
                 name=f"{name}_{i}",
-                data=points[i : i + 1, :].astype(np.float64),  # shape (1, 2)
-                confidence=np.ones(1, dtype=np.float64),
+                data=np.tile(points[i].astype(np.float64), (2, 1)),  # shape (2, 2)
```
gbeane, line 1019: **On-disk layout changed but `_JABS_NWB_FORMAT_VERSION` stays at 1.**

The static-object representation written into the file changes from `data` shape `(1, 2)` / `timestamps [0.0]` to `(2, 2)` / `[0.0, end_time]`, yet `_JABS_NWB_FORMAT_VERSION = 1` (line 43) is unchanged, so the `format_version` recorded in `jabs_metadata` is identical for files with two materially different static-object layouts.

Round-tripping is genuinely unaffected (`_read_static_objects` reads `series.data[0]`, so it handles both), so this is spec hygiene rather than a break. But `docs/development/jabs-nwb-format.md:435` documents `format_version` as "JABS NWB format version. Currently `1`" alongside the now-outdated `(1, 2)` field table, and line 540 refers to "future format versions" — an external consumer keying off `format_version` cannot tell which static-object shape to expect. Worth either bumping it or adding a note that version 1 covers both.

PR #464, opened 2026-09-10
Rebuild the feature manager when videos are pruned

gbeane:
## Reasoning

Follow-up to the unrelated finding Copilot raised while reviewing #463
([thread](https://github.com/KumarLabJax/JABS-behavior-classifier/pull/463#discussion_r3979332093)):
`FeatureManager` caches the project's capabilities and nothing rebuilds it when
videos are removed.

`FeatureManager` computes `_min_pose_version`, `_static_objects`,
`_distance_unit`, `_lixit_keypoints` and `_extended_features` once in
`__init__`, and `Project` builds it once at
[`project.py:222`](https://github.com/KumarLabJax/JABS-behavior-classifier/blob/2dbb5bc/src/jabs/project/project.py#L222-L227)
from the video set present when the project was opened. The GUI's *Prune
Project* action then removes videos from the live `VideoManager`
([`menu_handlers.py:426`](https://github.com/KumarLabJax/JABS-behavior-classifier/blob/2dbb5bc/src/jabs/ui/main_window/menu_handlers.py#L426))
without rebuilding it, so the project keeps reporting capabilities derived from
videos whose files were just deleted.

That is user-visible: `main_window.py` drives the Features menu off
`is_cm_unit`, `can_use_social_features`, `can_use_segmentation_features` and
`static_objects`, so after pruning the last v3 (or pixel-only, or
corners-less) video, the features those videos were blocking stay greyed out
until the project is closed and reopened. `get_project_defaults()` — the
fallback settings for behaviors with no saved settings — reads the same stale
values.

The staleness predates #463 and spans `Project`, `FeatureManager` and
`MainWindow`, which is why it did not belong in that PR.

**Two decisions this needed, both settled the same way — the live project
should match what reopening it produces:**

1. *Should pruning the last pixel-only video promote the project to cm units?*
   Yes. Reopening already does exactly that; leaving the live project behind
   is the bug, not the promotion.
2. *Should the refreshed defaults be written back to `project.json`?* Yes, and
   only when they change — the same condition `Project.__init__` already
   applies on every open, so this only moves that write earlier.

**Why it's safe:** removing a video can only relax the constraints
`FeatureManager` derives — `min_pose_version` is a `min()`, `static_objects` an
intersection, cm units require *all* videos to have a scale, `lixit_keypoints`
is a `min()` (three-keypoint lixit is backwards compatible with one). So a
refresh can only gain a capability, never take one away; no already-enabled
feature or saved per-behavior setting can be invalidated by it. No computed
feature values change, so no `FEATURE_VERSION` bump.

### Logic changes

**`src/jabs/project/project.py`**
- Retain the project scan results as `self._scan_results` (previously a local
  in `__init__`). These are small per-video dicts, and holding them is what
  lets a rebuild avoid re-running `_run_video_scan()`, which spins up a process
  pool and re-reads every pose file.
- New public `refresh_feature_manager()`: rebuilds `FeatureManager` from
  `video_manager.videos` and the retained scan results, then persists the
  refreshed defaults. Documented as removal-only — the current videos must be a
  subset of those scanned at open; added videos still require reopening the
  project (nothing adds videos to a live project today).
- Extract the defaults write from `__init__` into `_save_project_defaults()`,
  shared by `__init__` and the refresh (it also drops a duplicate
  `get_project_defaults()` call on the `__init__` path).

**`src/jabs/ui/main_window/main_window.py`** — extract the feature-availability
block out of `_project_loaded_callback()` into a public
`update_feature_availability_menus()` so it can be re-run after a prune. Same
behavior; the landmark loop collapses to `setEnabled(obj in available_objects)`.

**`src/jabs/ui/main_window/menu_handlers.py`** — the prune handler calls
`refresh_feature_manager()` and then `update_feature_availability_menus()`
after removing the videos, before refreshing the video list.

The CLI prune path (`jabs-cli prune`) needs nothing: it exits after pruning and
never reads `feature_manager` again.

### Mechanical updates

None.

## Verification

`uv run ruff check .`, `uv run ruff format .`, and `uv run pytest` (1141 passed)
all clean. No files under `packages/` were touched.

New tests:

**`tests/project/test_project.py`** — a two-video fixture (a v3 pose file with
no cm scale and no static objects, a v6 pose file with both), so the v3 video
holds every capability back:
- `test_refresh_feature_manager_recomputes_capabilities`: after removing the v3
  video, `min_pose_version` 3 → 6, segmentation and cm units become available,
  and `static_objects` becomes `{"corners"}`.
- `test_refresh_feature_manager_matches_reopening_the_project`: the refreshed
  capabilities, extended features and defaults equal those of a freshly opened
  project — the property this change is built on.
- `test_refresh_feature_manager_persists_updated_defaults`: the `defaults`
  block in `project.json` follows the new video set.
- `test_refresh_feature_manager_reads_no_pose_files`: deletes every pose file
  before refreshing and still gets the right capabilities, which it could not
  if the rebuild opened them.

**`tests/ui/test_main_window.py`** — `update_feature_availability_menus()`
enables exactly the supported features (and only the available landmarks), and
disables everything for a project that supports nothing extra.

**`tests/ui/test_menu_handlers.py`** — pruning removes the videos, then
refreshes the feature manager, then updates the menus, in that order; the
"would remove all videos" guard short-circuits before any of it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gbeane:
Copilot's finding (the suppressed comment on `menu_handlers.py:433`) is addressed: it reviewed this branch before #463 landed, and `main` has since been merged in at 734341a. `VideoManager.remove_video()` on this branch now decrements `_total_project_identities`, drops the `_video_has_cm_per_pixel` and `_pose_path_cache` entries, and removes the video's `video_files` entry from `project.json`, so the refresh call is no longer relying on a fix that isn't here.

The two changes also compose cleanly: `remove_video()` pops the pruned video's pose-path cache entry, and `refresh_feature_manager()` looks up cached pose paths only for the videos that remain.

Post-merge on the combined tree: `uv run pytest` 1144 passed, ruff clean. GitHub reports the PR mergeable with a clean merge state.

PR #469, opened 2026-09-14
KLAUS-660 Draw the last cv2 annotations as player overlays

gbeane:
The movement track, the closest-animal markers and the arena landmarks were the last three annotations still drawn into the frame by cv2 as the video was decoded. They are now overlays in `ui/player_widget/overlays`, painted over the frame, with their drawing shared through `jabs.overlay_drawing` alongside the pose, label and segmentation overlays.

Finishes the migration started in #467, which is where this was raised: segmentation became an overlay and so began painting over these three, inverting the order the decoder had composited them in.

## Paint order

Now set deliberately: **track, segmentation, closest-animal markers, landmarks**, then the overlays that were always Qt. That is how `_prepare_image()` composited them before segmentation moved, so the contours are back under the markers and the landmarks rather than over them.

## Three things that fall out of the move

- **`video_reader` is a pure reader.** `frame_annotation.py` is deleted, and `__scale_annotation_size()` with it - the overlays scale from base sizes by the zoom, the way the existing ones do. The package's Qt-free constraint still matters (`jabs.project.parallel_workers` imports it) but there is no drawing left to keep out.
- **`PlayerThread` is a pure decoder.** It no longer takes a pose object, an identity, or a list of identities, and its overlay flags, signals and slots are gone.
- **Selecting a different identity repaints instead of re-decoding the frame.** The track and the closest-animal markers were the only baked annotations that depended on the selection, so `set_active_identity()` now repaints and the seek is dropped.

## The closest-animal measurement

The ticket flagged this as the one part that is not just drawing: `PlayerThread._get_closest_animal_id()` measured hull distances and view angles across every identity, and it ran off the GUI thread.

It moved to `closest_identity.py` rather than into the overlay, and the overlay caches the result per frame and active identity. A repaint can be triggered by things that change neither - a resize, a crop, another overlay toggling - and those must not redo the measurement. There is a test asserting a second identical repaint does not re-measure.

## Appearance

I rendered the same scene through the old cv2 functions and the new overlays:

| | old (cv2) | new (Qt) |
|---|---|---|
| painted pixels | 1594 | 1062 |
| bounding box | x 17-383, y 18-383 | x 18-381, y 18-381 |
| corner / lixit / hopper / future / past colors | all present | all present |

Geometry and colors are unchanged. The new drawing paints about a third fewer pixels because cv2 drew its tiny circles with `LINE_AA`, whose soft edge is wider than Qt's - the same class of difference as the antialiasing note in #467, and visually the two renders are indistinguishable.

## Also

- Drops the unreachable centroid-tracking mode of `draw_track()` - the player only ever asked for a nose track.
- Adds the return type hint to `PoseEstimation.fps`, too small to be worth its own PR.
- Documents nothing new in the user guide: the menu items, the `?` key and what they draw are all unchanged.

## Testing

`uv run pytest tests` - 1275 passed (33 new); with `PySide6` made unimportable (how CI sees it) 894 passed, 381 skipped, no collection errors; `packages/jabs-core/tests` 97 passed; ruff clean.

New tests cover the track drawing (past and future in different colors, anchored on the current frame, a masked keypoint breaking the line rather than being bridged, crop clipping, and frame 0 not wrapping around to the end), the landmark drawing (nothing drawn without static objects, corners, one- and three-keypoint lixit, a closed hopper outline, crop clipping), the closest-identity measurement (nearest animal, field of view excluding what is behind the subject, missing poses, a subject without nose and neck keypoints), and the overlays themselves (off by default, repaint on toggle, track limited to the active identity, both markers drawn only when they differ, the per-frame measurement cache, and the paint order).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gbeane:
Implements KLAUS-660.

src/jabs/ui/player_widget/player_widget.py
```diff
@@ -78,10 +78,6 @@ def __init__(self, *args, **kwargs) -> None:
         self._pose_est = None
         self._playback_range: PlaybackRange | None = None
 
-        # properties to control video overlays managed by PlayerThread
-        self._label_closest = False
-        self._show_track = False
-        self._overlay_landmarks = False
         self._identities = []
```
gbeane: **`self._identities` is now write-only — the last reader went away with this PR.**

It is assigned here, at line 241 (`= None`) and line 274 (`= pose_est.identities`), and read nowhere: `git grep 'self\._identities' -- src/jabs/ui` returns only those three writes. Its one reader was the `identities` argument to `PlayerThread`, which this PR drops from the constructor call at line 276.

Worth noting the three sibling flags in this same hunk (`_label_closest`, `_show_track`, `_overlay_landmarks`) *were* removed — this one looks simply missed.

Also inconsistent while it survives: `[]` here vs `None` at line 241, which nothing can catch any more now that there is no reader.

src/jabs/overlay_drawing/landmarks.py
```diff
@@ -0,0 +1,101 @@
...
+            ]
+        else:
+            # Older pose files record a lixit as a single keypoint.
+            points = [(lixit[i, 0], lixit[i, 1]) for i in range(lixit.shape[0])]
+        _draw_points(painter, points, to_output, LANDMARK_LIXIT_COLOR, point_radius)
+
+    hopper = static_objects.get("food_hopper")
+    if hopper is not None:
+        mapped = [to_output(float(x), float(y)) for x, y in hopper]
+        if all(point is not None for point in mapped):
```
gbeane: **`all(...)` hides the entire food hopper when any one corner is cropped out.**

`to_output` is `image_to_widget_coords_cropped`, which returns `None` for any point outside `crop_rect` (`frame_with_overlays.py:420`). So zooming in far enough that one of the four hopper corners leaves the view drops the whole polygon rather than clipping it. The hopper spans a good part of the arena, so this is reachable on any meaningful zoom.

This is a regression against the cv2 code being replaced: `frame_annotation.py:180` on `main` called `cv2.polylines(img, np.int32([hopper_points]), True, ...)` unconditionally, and cv2 clips to the image.

The pattern for this already exists and is used by the sibling closed-shape path — `segmentation.py:121-134` maps, then walks `visible_runs(mapped)` and picks `drawPolygon` when `complete` else `drawPolyline`. `visible_runs`' docstring calls out that `complete` is there precisely for "callers that close a shape". The hopper is a closed shape and is the one such caller not using it.

src/jabs/ui/player_widget/overlays/closest_identity_overlay.py
```diff
@@ -0,0 +1,100 @@
...
+            return
+
+        zoom = self.parent.scaled_pix_width / max(crop_rect.width(), 1)
+        radius = max(1, round(CLOSEST_MARKER_RADIUS * zoom))
+
+        # The in-view animal is drawn second so it stays visible when one animal is
+        # both the nearest overall and the nearest in view.
+        for identity, color in (
+            (closest_id, CLOSEST_MARKER_COLOR),
+            (closest_fov_id, CLOSEST_FOV_MARKER_COLOR),
```
gbeane: **The comment's justification describes a state `_closest_identities` rules out.**

The comment says the in-view animal is drawn second "so it stays visible when one animal is both the nearest overall and the nearest in view" — but line 96-97 does:

```python
if closest_id == closest_fov_id:
    closest_id = None
```

so the two are never the same identity, and the `identity is None` guard skips the one that was cleared. The case the comment justifies cannot happen.

Separately, the order is inverted relative to the cv2 code this replaces: `player_thread.py:128-145` on `main` drew the FOV marker first and the overall-closest second, with the same `closest_id != closest_fov_id` guard. Since the two markers are always different animals, the flip only changes which one wins if two centroids happen to land on overlapping pixels — so this is about the comment being misleading more than the pixels, but the order change also looks unintended.

gbeane (commented):
Review of the overlay migration (low effort level, so a short list). All three verified against the branch: one behavior regression on zoom, one leftover from the `PlayerThread` teardown, and one comment whose stated reason cannot occur.

src/jabs/ui/player_widget/player_widget.py
gbeane: Fixed in cdcdb1ea - removed all three writes. Confirmed it was write-only: `git grep 'self\._identities' -- src/jabs/ui` returns only those lines, and the `identities` argument this branch dropped from the PlayerThread constructor was indeed its last reader. You are right that it was missed rather than kept deliberately.

One neighbour worth knowing about, which I have left alone: `update_identities` (declared at player_widget.py:58, documented at :41) is never emitted and nothing connects to it - `git grep update_identities` finds only those two lines. That one predates this branch, so I did not fold its removal in. Say the word and I will, either here or on its own.

src/jabs/ui/player_widget/overlays/closest_identity_overlay.py
gbeane: Fixed in cdcdb1ea, both halves.

The comment was justifying a state `_closest_identities()` rules out, exactly as you describe. And the order was flipped relative to `player_thread.py` on main, which drew the FOV marker first. Restored to in-view first, then nearest overall, and the comment now says the real reason: the two are always different animals, so the order only decides which marker wins where two centroids land on overlapping pixels. The existing both-markers test now asserts the order too.

src/jabs/overlay_drawing/landmarks.py
gbeane: Fixed in cdcdb1ea - and this one was worse than the report, in a way the report led me straight to.

Switched the hopper to `visible_runs(mapped)` as suggested, `drawPolygon` when `complete` else `drawPolyline`, matching the segmentation path. Then the regression test I wrote for it failed: with the hopper's two right-hand corners cropped out, nothing was drawn at all.

The reason is in `visible_runs` itself, and the segmentation contours had it too. It walked the points as a flat list, so for a closed shape the run spanning the wrap-around edge came out as two separate runs - the head and the tail of the list - and the edge between the last point and the first was lost. In the hopper case the two surviving corners are adjacent *only* through that edge, so both runs were single points and both were skipped.

`visible_runs` now takes `closed=True` and walks the points as a ring, rotating the sequence to begin at a gap so a wrap-around run stays in one piece. Both closed-shape callers pass it. Tests added for the hopper and for a contour clipped across its wrap-around; both fail with the rotation disabled.

Worth stating the limit of the approach, since it is not what cv2 did: clipping happens at point granularity, not pixel granularity, so an edge that crosses the crop boundary is dropped rather than truncated at the boundary. Near a crop edge the outline stops at the last visible corner instead of running to the edge of the view. That is the same tradeoff the segmentation contours already make, and it beats dropping the shape, but if you want true clipping the fix is a clip region on the painter rather than a `to_output` that returns None.

src/jabs/ui/player_widget/player_widget.py
gbeane: Removed in 71c3cb29 - the declaration and its Signals docstring entry are both gone, and `git grep update_identities` is now empty.

Note the branch tip was force-pushed: my first attempt at this commit (df01102f) accidentally included `.claude/worktrees/annotation-store-seam` as a gitlink, because `git add -A` swept up another session's worktree, which is not gitignored. Amended it away and force-pushed with --force-with-lease; only that tip commit changed, and the diff against main is now src/, tests/ and packages/ only.

gbeane:
Stacked #471 on this branch: it drops the past track markers from radius 2 to radius 1 so both halves of the track are the same size. Base is `feature/klaus-660-overlay-migration`, so it will need retargeting to main if this merges first.

PR #472, opened 2026-09-14
KLAUS-681 move annotation JSON reads and writes behind an AnnotationStore

gbeane:
Foundation work for Hub-backed JABS projects — Phase 0 of `docs/development/plans/jabs-hub-backed-projects-plan.md` (§4.4b). Pure refactor: local-only behavior is unchanged.

Today every read and write of a project's behavior labels (`jabs/annotations/<video>.json`) is concrete `pathlib` + `json`, spread across `Project`, `VideoManager`, `project_pruning`, and the feature-extraction workers. A Hub-backed project needs those documents to come from a cache that syncs with JABS Hub, so the access has to go through one interface first.

## What this adds

New `jabs.io.annotations` package (`packages/jabs-io/src/jabs/io/annotations/`):

- **`AnnotationStore`** — ABC over the *serialized document dict*, not `VideoLabels`. Building a `VideoLabels` requires pose data, so serialization stays in `src/jabs/project/video_labels.py`; keeping the store at the dict level is what lets it live in `jabs-io` at all.
- **`AnnotationDocument`** — `NamedTuple` of `(content, version)`, so it unpacks like the plan's `tuple[dict, int]` but reads as named fields at the call sites.
- **`LocalAnnotationStore`** — today's behavior exactly: one JSON file per labeled video, written atomically (temp file + rename). A project directory has no version history, so it reports every document at the `UNVERSIONED` sentinel and ignores `base_version`.
- **`read_document` / `write_document`** — path-level helpers for the worker processes.

Interface: `load_document`, `save_document`, `has_document`, `document_path`, `ensure_local`.

`base_version` is carried from the start, unused by the local store, so adding optimistic concurrency in Phase 3 doesn't re-touch every call site.

## Routed through `Project.annotation_store` (shared with `VideoManager`)

- `Project.save_annotations` — the single write seam
- `VideoManager.load_video_labels` / `load_annotations` / `annotations_path`
- `Project.load_counts`, `Project.labeled_identities`, `Project.get_derived_file_paths`
- `project_pruning.get_videos_to_prune`
- `Project._build_feature_load_job_base` — calls `ensure_local()` so the child process still reads by `Path` and stays network-free; `parallel_workers._load_video_labels` parses with `read_document` rather than its own `json.load`

## Deliberately out of scope

Recorded in a new §4.4.1 of the plan, with the reason each was left and what would change it — so they get revisited deliberately rather than rediscovered. These manipulate annotation *files and directories* rather than *documents*, so routing them through the store would push local-filesystem operations (copy the directory, back it up, swap it into place) into an interface that has to work against Hub:

- `scripts/cli/update_pose.py` and `update_labels.py` staged-rewrite machinery. Pose upgrade is expected to move server-side for Hub-backed projects (plan §4.6, D20), which makes the staged local rewrite the *local* project's code path rather than something to abstract.
- `classifier/mlflow_logging.py:archive_annotations` — zips the directory as an opaque provenance artifact; never reads a document.
- `_orphan_identities` and `_source_annotated_videos` are genuine document reads and the cheapest follow-ups to pick up, since `LocalAnnotationStore` can already be constructed over an arbitrary directory.

## Deviations from the plan's protocol sketch

- **Added `has_document` / `document_path`** for callers that only need existence or a location and must not trigger a fetch (`get_derived_file_paths`, `project_pruning`). The sketch had no way to express that without paying for a download.
- **Dropped `list_labeled_videos`** — no in-scope caller needs it, and enumerating a directory versus a manifest are different enough operations to design alongside the Hub store rather than ahead of it.

## Tests

45 new:

- `packages/jabs-io/tests/annotations/test_store_contract.py` — implementation-agnostic expectations, parametrized over backends via the `store` fixture, so a Hub store is covered by adding one line rather than writing a second suite.
- `packages/jabs-io/tests/annotations/test_local.py` — local specifics: atomic write leaves no temp file, name derivation, directory creation, corrupt-JSON surfacing.
- `tests/project/test_annotation_store_seam.py` — substitutes an in-memory store into a real `Project` and asserts each call site consults it. If anything reverts to a direct read, a test fails instead of the drift going unnoticed.

Root suite 1246 → 1263 passed, `jabs-io` 299 → 327, `jabs-core` / `jabs-behavior` / `jabs-vision` unchanged, `ruff check` and `ruff format` clean.

## Docs

- Plan: §3.2 status, §4.4b rewritten to the landed interface, new §4.4.1, phasing table, risks table, Appendix A.3.
- `docs/development/development.md`: new "Behavior Annotation Storage" section covering the two rules for new code — go through the store, and worker processes get a path rather than a store.

## Review notes

- No behavior change is intended anywhere. The document filename derivation is identical to what each call site computed before (verified for dotted names, missing suffixes, and full paths).
- `LocalAnnotationStore.save_document` creates missing parent directories, which `Project.save_annotations` did not. `ProjectPaths.create_directories` already makes the directory, so this is a no-op in practice and only makes the store usable standalone.

Jira: KLAUS-681


packages/jabs-io/src/jabs/io/annotations/base.py
```diff
@@ -0,0 +1,147 @@
...
+    @abstractmethod
+    def has_document(self, video_name: str) -> bool:
+        """Return whether the store holds an annotation document for a video.
+
+        Args:
+            video_name: Video filename to check.
+        """
+
+    @abstractmethod
+    def document_path(self, video_name: str) -> Path:
```
gbeane, line 135: **The store has no delete operation, but two call sites take a path from it in order to delete the file.**

`project_pruning.py:36` gets `project.annotation_store.document_path(video)` into `VideoPaths.annotation_path`, and `ui/main_window/menu_handlers.py:452` folds that into `files_to_delete` → recycle bin / unlink. `Project.get_derived_file_paths` returns `document_path` in the same spirit.

This method's own docstring is what makes it a problem: the path is where the document lives "(for a local project) or would be cached (for a remote one)". So for any non-local store, pruning a video deletes the *cache entry* and leaves the authoritative document intact — and re-syncable, so the annotation reappears.

No bug today (the only store is local), but this is the one place the seam is incomplete in a way that changes behavior in Phase 3 rather than just needing more code. Either add `delete_document` now, or record it in the new §4.4.1 next to the other deferrals so it isn't discovered by a reappearing annotation.

packages/jabs-io/src/jabs/io/annotations/local.py
```diff
@@ -0,0 +1,155 @@
...
+
+    Args:
+        path: Path to write the annotation JSON file to.
+        document: Serialized document, as produced by ``VideoLabels.as_dict()``.
+
+    Raises:
+        OSError: If the file cannot be written.
+    """
+    path.parent.mkdir(parents=True, exist_ok=True)
+    tmp = path.with_suffix(".json.tmp")
```
gbeane: **The temp file name is deterministic, so concurrent writers can interleave into one document.**

`path.with_suffix(".json.tmp")` means every writer of a given video targets the same `<video>.json.tmp`. Two processes against the same project directory — a GUI session plus `jabs-cli merge` or `update_labels`, or simply two GUI instances — can both open, write and rename it, and the interleaving can publish a document that is a *mixture* of the two writes rather than last-writer-wins.

Pre-existing behavior, not introduced here. Worth raising anyway because this PR makes it a single chokepoint: a pid- or `mkstemp`-suffixed sibling fixes it in one place, for every caller, permanently.

packages/jabs-io/src/jabs/io/annotations/local.py
```diff
@@ -0,0 +1,155 @@
...
+        return None
+    with path.open() as f:
+        return json.load(f)
+
+
+def write_document(path: Path, document: AnnotationDocumentContent) -> None:
+    """Write a serialized annotation document to a JSON file atomically.
+
+    The document is written to a sibling temporary file and then renamed over
+    the destination, so a crash mid-write cannot leave a truncated annotation
```
gbeane: **The atomicity claim is stronger than the code delivers.**

"a crash mid-write cannot leave a truncated annotation file behind" holds for a *process* crash, but there's no `f.flush()` / `os.fsync(f.fileno())` before `tmp.replace(path)` on line 62. On power loss or a kernel panic the rename metadata can reach disk while the data blocks have not, leaving a zero-length `video1.json` where a valid document used to be — i.e. exactly the truncated file the docstring promises is impossible, and the old in-place write would at least sometimes have survived.

Either add the fsync or narrow the wording to process crashes.

src/jabs/project/video_manager.py
```diff
@@ -51,9 +55,15 @@ def __init__(
         enable_video_check: bool = False,
         *,
         scan_results: "dict[str, VideoScanResult]",
+        annotation_store: AnnotationStore | None = None,
```
gbeane: **Defaulting to `None` silently constructs a second store, which undercuts the invariant this PR establishes.**

When omitted, lines 62-66 build a fresh `LocalAnnotationStore(paths.annotations_dir)`. No live bug — `project.py:227` is the only construction in `src/` and it threads the store through. But the point of this refactor is one store per project, and this default means a future Hub-backed path that forgets to pass it reads local JSON files and never signals that anything is wrong.

Making the parameter keyword-only with no default makes that mistake unrepresentable. The cost is updating the test call sites that currently rely on the default (`tests/project/test_parallel_workers.py:321,354,385` and `tests/project/test_video_manager.py`).

packages/jabs-io/src/jabs/io/annotations/base.py
```diff
@@ -0,0 +1,147 @@
...
+            The document and its version, or ``None`` if the store holds no
+            document for this video.
+        """
+
+    @abstractmethod
+    def save_document(
+        self,
+        video_name: str,
+        document: AnnotationDocumentContent,
+        base_version: int | None = None,
```
gbeane, line 91: **The signature is future-proof but the data flow isn't — no caller can supply this yet.**

The PR body says carrying `base_version` means "adding optimistic concurrency in Phase 3 doesn't re-touch every call site", which is true of the *signature*. But nothing retains a version to pass: `base_version` appears only here and in `local.py:135` (both as the parameter itself), and all three `load_document()` callers in `src/` take `.content` and drop `.version` — `project.py:1707` (`load_counts`), `video_manager.py:156` (`load_annotations`), `video_manager.py:317`.

`Project.archive_behavior` (line 1069) is the concrete case: it's a read-modify-write over `load_video_labels` → `save_annotations`, exactly where a version has to flow, and today it can't.

Informational — no change needed in this PR. Worth stating in the phase notes so Phase 3 expects to thread versions through the call sites rather than assuming the plumbing is already done.

gbeane (commented):
Review of the annotation store seam (medium effort). No correctness regressions — I traced every moved call site and `annotation_filename()` reproduces what each one computed before, `load_annotations` still rejects out-of-project videos, and no stray annotation `json.load`/`dump` survives outside the three you declared out of scope. `pytest packages/jabs-io/tests/annotations tests/project` passes (404) and ruff is clean.

Five findings, all lower-severity; the delete gap is the only one I'd call a decision rather than a tidy-up.

packages/jabs-io/src/jabs/io/annotations/base.py
gbeane: Added `delete_document` to the interface, and recorded the call-site migration in §4.4.1 rather than doing it here.

The method is on the ABC because you are right that this is a decision, not a tidy-up — deletion has to reach the authority, and `LocalAnnotationStore.delete_document` is a plain `unlink` that returns whether anything was there. Contract tests cover it for every backend.

I did **not** rewire `project_pruning` / `menu_handlers` onto it, because the GUI path sends those paths to the OS recycle bin with an unlink fallback. Switching annotations to `delete_document` would silently drop that safety net for one file type — a behavior change, which this PR is not. So §4.4.1 now carries it as the one deferral that would surface as a bug rather than as missing code, naming the failure mode (prune recycles the cache entry, the next sync brings the labels back) and flagging it as a blocker for any non-local store.

packages/jabs-io/src/jabs/io/annotations/local.py
gbeane: Fixed. The temporary is now `<video>.json.<pid>.<8 hex>.tmp`, so each writer stages through its own file and the rename decides.

Two details worth flagging:

- I used a normal create rather than `tempfile.mkstemp`, deliberately — `mkstemp` forces `0600`, which would make annotation files unreadable to other members of a lab sharing a project directory over NFS. A plain create keeps the umask behavior this has always had. There is a comment on it so nobody "improves" it to `mkstemp` later.
- Unique names mean a failed write no longer overwrites one stale `.tmp` but accumulates a new file each time, so the write is wrapped in `try/except BaseException` with an `unlink(missing_ok=True)`.

Covered by `test_concurrent_writers_do_not_share_a_temporary_file`, which re-enters `save_document` while the first write holds its temporary open. I checked it against the old implementation: it reuses `video1.json.tmp` for both writes, so the distinct-temporaries assertion fails. Also added `test_a_failed_write_leaves_no_temporary_behind` and `test_a_failed_write_leaves_the_previous_document_intact`.

packages/jabs-io/src/jabs/io/annotations/local.py
gbeane: You are right, and I narrowed the wording rather than adding the fsync.

The reason for choosing that direction: the GUI saves this file synchronously on **every** label edit (`central_widget.py`, several call sites). An `fsync` per save puts a disk round-trip directly in the painting path, which is a real interaction cost paid on every stroke to buy durability against power loss — and the old in-place write did not buy that either, so adding it here would be a behavior change smuggled into a refactor.

The docstring now says what the code does: a *process* crash cannot corrupt the document, but a power loss or kernel panic can lose the most recent save because the rename may reach disk before the data blocks. It also states why the fsync is absent, so the trade is visible to whoever reconsiders it rather than reading as an oversight. Same correction applied to the developer docs.

If durability there turns out to matter, the honest version is probably fsync on an explicit save/close rather than on every edit — worth its own ticket.

src/jabs/project/video_manager.py
gbeane: Agreed — made it keyword-only with no default, so the mistake is unrepresentable.

I had defaulted it to keep the test churn down, which was the wrong trade for exactly the reason you give: the whole point of the refactor is one store per project, and a silent second one fails in the quietest possible way.

Updated five call sites (`test_parallel_workers.py` ×3, `test_video_manager.py` ×2) to pass `LocalAnnotationStore(paths.annotations_dir)`. The seam test that asserted the default is replaced by two: one that the constructor now raises `TypeError` without a store, and one that the store passed in is the one the manager reports.

packages/jabs-io/src/jabs/io/annotations/base.py
gbeane: Fair catch — my PR-body claim was too strong, and I corrected the plan rather than leaving it to mislead Phase 3.

§4.4b now reads: carrying `base_version` future-proofs **the signature, not the data flow**. It names the gap you identified — all three `load_document()` callers take `.content` and drop `.version` — and calls out `Project.archive_behavior` as the concrete read-modify-write where a version has to flow and currently cannot, so Phase 3 budgets for threading versions through the call sites instead of assuming the plumbing is done.

No code change, as you said.

PR #475, opened 2026-09-16
Show manual labels and predictions side by side in the player and video export

gbeane:
## Summary

The label overlay could show either the manual label or the prediction. It can now show both at once, beside each animal, in the player and in an exported video.

- **Player:** `View → Label Overlay` gains **Labels and Predictions**. Each animal gets two markers, the manual label first and the prediction next to it. The pair is placed as a unit, so it still keeps clear of the identity label in every identity overlay mode.
- **Video export:** the options dialog gains a **Manual labels** checkbox alongside **Behavior predictions**. They are independent, so an export can carry either or both, and the burned-in caption and legend name whichever markers were chosen:

  | chosen | caption | gray swatch |
  |---|---|---|
  | labels | `Grooming labels` | `no label` |
  | predictions | `Grooming predictions (raw)` | `no prediction` |
  | both | `Grooming labels (left) and predictions (raw, right)` | `no label or prediction` |

Binary and multi-class projects both work: in multi-class mode the labels and the predictions are indices into the same timeline color table, so one table and one legend cover the pair.

## Notes for review

- **`PredictionOverlay` → `LabelMarkerOverlay`** (`jabs/video_export/label_marker_overlay.py`). It now carries `manual_labels` and `predicted_labels`, exposes `marker_count` / `marker_values()`, and refuses construction with no source rather than burning a caption into every frame for markers that never appear. The `prediction_overlay=` keyword on `render_overlay_frame`, `export_overlay_video` and `VideoExportThread` is now `label_overlay=`.
- **`CentralWidget.prediction_overlay()` becomes `export_label_markers()`**, returning an `ExportLabelMarkers` value object carrying both sources' arrays and the reasons a source cannot be drawn. The handler gathers once before the options dialog, for the tooltips, then calls `markers.overlay(labels=, predictions=)` after it, since the caption depends on the choice - one gather and one build, as before this change. The manual labels come from the same `_manual_overlay_labels()` the player uses, so the export and the player agree by construction.
- **A stale multi-class prediction record now only disables predictions.** The labels are built from the project's current behavior list, so they cannot go stale with the saved record and stay exportable on their own.
- **Manual labels are always offered** for a loaded video, labeled or not; unlabeled frames get the gray marker, exactly as the player draws them.
- **The three copies of the label-overlay refresh logic in `CentralWidget`** (mode setter, `_set_label_track`, `_set_prediction_vis`) collapse into one `_refresh_label_overlay()`. It skips pushing an already-empty overlay, so labeling with the overlay switched off no longer reloads the displayed frame a *second* time - `_label_button_common` still reloads it once regardless, which this deliberately leaves alone.
- **The marker gaps moved into `jabs.overlay_drawing`** as `LABEL_MARKER_PAIR_GAP`, with `native_label_marker_sizes()` returning `(marker_size, gap, pair_gap)`. The player now uses the shared constants instead of its own copies - that package exists so the player and the exports cannot drift apart.

## Review follow-ups (644f6f0c)

Five findings from the first review, all addressed:

- Empty label sources (a pose file with no identities) are reported as unavailable by the `_export_*` helpers instead of reaching `LabelMarkerOverlay.__post_init__` and raising out of the export handler.
- `_draws_a_marker()` now drives the `sources` property, the captions and the legends alike, so a present-but-empty source can no longer be named in a caption for a marker that is never drawn.
- The two-methods API became one `export_label_markers()` gather, so a multi-class export does one pass over the labels and predictions rather than two.
- The player's marker loop bounds-checks the frame as well as the identity, matching the exporter's `_value`: in `BOTH` mode the two sources can run out on different frames.
- Corrected the overstated claim about frame reloads (above).

## Testing

- `uv run pytest` - 1364 passed, 0 skipped. (The export test modules had been silently skipping on the renamed import until they were updated.)
- `uv run ruff check . && uv run ruff format --check .` - clean.
- Exported a real 10-frame video offscreen with both markers and checked the caption, the 3-entry legend, the two markers and the written mp4.
- Built the real menus and the real `CentralWidget` + options dialog offscreen: all four `View → Label Overlay` entries map to their mode, and the four export checkboxes come up with the right enabled state and tooltips.

User-guide docs updated in both copies (`docs/user-guide/` and `src/jabs/resources/docs/user_guide/`).


src/jabs/ui/main_window/central_widget.py
```diff
@@ -1554,28 +1600,107 @@ def _showing_postprocessed_predictions(self) -> bool:
...
-            ``(overlay, unavailable_reason)``, exactly one of which is set. The reason
-            is UI copy for the disabled checkbox's tooltip, and comes from here rather
-            than being inferred by the caller so that it cannot disagree with the
-            decision it explains.
+            The overlay, or ``None`` when neither source was asked for or neither is
+            available. :meth:`label_markers_unavailable` says why a source is not.
+        """
+        manual_labels = self._export_manual_labels()[0] if draw_labels else None
+        predicted_labels = self._export_predicted_labels()[0] if draw_predictions else None
+        if manual_labels is None and predicted_labels is None:
```
gbeane: This guards on `None`, but `LabelMarkerOverlay.__post_init__` rejects sources that are *empty sequences* too - `sources` filters on `len(source) > 0`, not on `is not None`.

Every producer here returns a list sized `num_identities`, so a pose file that yields zero identities makes `_export_manual_labels()` / `_export_predicted_labels()` return `([], None)`. That reads as "available", so `label_markers_unavailable()` returns no reason and the checkbox comes up enabled; the user ticks it, and this method then falls through the `is None` guard and constructs the overlay, which raises `ValueError("a label marker overlay needs manual labels, predictions, or both")` straight out of `MenuHandlers.export_overlay_video` - there is no `try/except` anywhere on that path.

Guarding on emptiness here (`if not manual_labels and not predicted_labels:`), or having the `_export_*` helpers normalize `[]` to `(None, reason)`, closes it.

src/jabs/video_export/label_marker_overlay.py
```diff
@@ -0,0 +1,250 @@
...
+            ValueError: If neither source was given, which would burn in a caption for
+                markers that are never drawn.
+        """
+        return cls(
+            manual_labels=manual_labels,
+            predicted_labels=predicted_labels,
+            color_lut=None,
+            caption=_caption(
+                subject=behavior,
+                labels=manual_labels is not None,
```
gbeane: Same `None`-vs-empty split as in `CentralWidget.label_marker_overlay`, but internal to the class: `_caption()` and `_no_value_label()` are fed `manual_labels is not None` / `predicted_labels is not None`, while `marker_count` and `marker_values()` are driven by the emptiness-filtered `sources` property.

A source that is present but empty therefore burns in `"Grooming labels (left) and predictions (raw, right)"` and the `"no label or prediction"` legend entry while only one marker is ever drawn - so the caption names the wrong square. Deriving the caption/legend flags from the same emptiness test the drawing uses would keep them in step by construction, which is what the class docstring promises the constructors do.

src/jabs/ui/main_window/menu_handlers.py
```diff
@@ -223,32 +224,41 @@ def export_overlay_video(self) -> None:
             )
             segmentation_unavailable = f"No segmentation data available: {reason}"
 
-        # Built up front so the checkbox can say whether there is anything to draw,
+        # Asked up front so each checkbox can say whether there is anything to draw,
         # and why not when there is not.
-        prediction_overlay, predictions_unavailable = central_widget.prediction_overlay()
+        labels_unavailable, predictions_unavailable = central_widget.label_markers_unavailable()
```
gbeane: Performance regression worth a look: `label_markers_unavailable()` here and `central_widget.label_marker_overlay(...)` after the dialog both go through `_export_manual_labels()` / `_export_predicted_labels()`, and those do the *full* work rather than a cheap availability check - `_build_multiclass_overlay_labels()` allocates `num_identities x num_frames` arrays, `build_multiclass_label_array` walks every behavior, and `_get_prediction_list()` rebuilds the whole prediction list.

Before this change the overlay was built exactly once. Now a multi-class export does two full builds of both sources, and the first one runs synchronously *before* the options dialog is even shown, so on a long video the UI stalls on a menu click with nothing on screen to explain it. Caching the result of the availability pass and reusing it after the dialog would restore the old cost while keeping the "caption depends on the choice" ordering.

src/jabs/ui/player_widget/overlays/label_overlay.py
```diff
@@ -54,19 +80,27 @@ def _overlay_labels(self, painter: QtGui.QPainter, crop_rect: QtCore.QRect) -> N
...
-            behavior_y = widget_y - LABEL_MARKER_SIZE
+            for position, values in enumerate(label_sources):
+                if identity >= len(values):
+                    # a label source can be short an identity if it was built for a different
+                    # pose file. skipping the marker keeps the rest of the group in place.
+                    continue
 
-            label_val = int(self.parent.labels[identity][self.parent.current_frame])
-            prediction_color = label_marker_color(label_val, self.parent.label_color_lut)
+                label_val = int(values[identity][self.parent.current_frame])
```
gbeane: The new loop added an `identity >= len(values)` guard but no frame guard, while its export counterpart `LabelMarkerOverlay._value` bounds-checks *both* identity and `frame_index` and documents why ("Labels and predictions can be shorter than the video").

In `BOTH` mode the two sources can have different per-identity lengths - manual labels are sized to `VideoLabels` / pose frames, predictions come from the saved prediction record - and the codebase already acknowledges video-vs-pose frame-count mismatch via `_frame_count_mismatch_message`. When `current_frame` exceeds one source's length this raises `IndexError` inside `paintEvent`.

Pre-existing for the single-source modes, but `BOTH` doubles the chance of hitting a short array, and the asymmetry with the exporter is worth closing here.

src/jabs/ui/main_window/central_widget.py
```diff
@@ -1514,32 +1471,121 @@ def _build_multiclass_overlay_labels(self) -> list[np.ndarray]:
...
+            return
+
+        self._player_widget.set_label_color_lut(color_lut)
+        self._player_widget.set_labels(manual_labels, predicted_labels)
+        self._label_overlay_populated = True
+
+    def _clear_label_overlay(self) -> None:
+        """Remove any values the label overlay is drawing.
+
+        Does nothing if the overlay has nothing to draw already, so that labeling with
```
gbeane, line 1630: This claim (and the matching line in the PR description) overstates what the short-circuit does: `_label_button_common` still calls `self._player_widget.reload_frame()` unconditionally, immediately before `self._set_label_track()`.

So with the overlay switched off an edit goes from two frame reloads (the explicit one plus the one inside `set_labels`) down to one, not to zero - the displayed frame is still reloaded on every edit. Worth rewording to "does not reload the displayed frame a second time", or dropping the explicit `reload_frame()` in `_label_button_common` if the intent really was zero.

gbeane (commented):
Correctness review of the label/prediction overlay change. The refactor is coherent: the geometry math in the player and the exporter agree (`group_width = n*size + (n-1)*pair_gap`, `group_x = x - group_width - gap`, `stride = size + pair_gap`), the rename is complete with no stale `PredictionOverlay` / `prediction_overlay=` references, and all three `native_label_marker_sizes` callers were updated for the new 3-tuple. `tests/ui tests/video_export tests/overlay_drawing` pass (445).

Five findings inline, one medium and four low. The medium one is a `None`-vs-empty-sequence mismatch between `label_marker_overlay()`'s guard and `LabelMarkerOverlay.__post_init__`'s validation.

Things I checked and found fine: the `_label_overlay_populated` flag is only ever set `True` alongside a real push and `False` alongside a real clear, so I could not construct a stale-marker case from it; `_showing_postprocessed_predictions` is safe on the new labels-only export path because both prediction dicts default to `{}`; and the multi-class LUT / `class_names` staleness check is unchanged from `main`.

src/jabs/ui/main_window/central_widget.py
gbeane: Fixed in 644f6f0c, by normalizing in the `_export_*` helpers as you suggested - guarding only at the construction site would have left the checkbox enabled with nothing behind it. Both now return `(None, reason)` for an empty list, so zero identities reads as "unavailable" rather than "available but empty", and `ExportLabelMarkers.overlay()` guards on emptiness too as insurance for a directly constructed one.

src/jabs/video_export/label_marker_overlay.py
gbeane: Fixed in 644f6f0c. Added `_draws_a_marker(source)` and routed the `sources` property, both captions and both legends through it, so the caption/legend flags and the drawing are the same test by construction. Covered by a test that an empty label source with real predictions captions itself `Grooming predictions (raw)` with the `no prediction` swatch and draws one marker.

src/jabs/ui/main_window/menu_handlers.py
gbeane: Fixed in 644f6f0c. `label_markers_unavailable()` and `label_marker_overlay()` are replaced by one `export_label_markers()` returning an `ExportLabelMarkers` value object that carries the gathered arrays plus the reasons; the handler gathers once before the dialog and calls `markers.overlay(labels=..., predictions=...)` after it. One gather, one build, so a multi-class export is back to the single pass it did before - and the pre-dialog pass is now the only one. A test asserts `export_label_markers` is called exactly once and records which markers `overlay()` was asked for.

src/jabs/ui/player_widget/overlays/label_overlay.py
gbeane: Fixed in 644f6f0c: the loop now bounds-checks the frame against the per-identity array before reading it, with a comment on why the two sources can run out on different frames. Test added for a manual label array that ends before the current frame - the prediction marker still draws, in the slot it would have had. I left the negative-index case out, since `current_frame` is never negative in the player, unlike the exporter where `_value` takes arbitrary ints.

src/jabs/ui/main_window/central_widget.py
gbeane: Right - reworded in 644f6f0c to "does not reload the displayed frame a second time", with a note that `_label_button_common` reloads once regardless. I did not drop that reload: with the overlay off nothing on the frame depends on the labels, so it looks removable, but that is a user-visible change to what an edit repaints and belongs in its own change rather than buried in this refactor. The PR description is corrected to match.

src/jabs/ui/main_window/central_widget.py
gbeane, line 1634: This guards on `None`, but `LabelMarkerOverlay.__post_init__` rejects sources that are *empty sequences* too - `sources` filters on `len(source) > 0`, not on `is not None`.

Every producer here returns a list sized `num_identities`, so a pose file that yields zero identities makes `_export_manual_labels()` / `_export_predicted_labels()` return `([], None)`. That reads as "available", so `label_markers_unavailable()` returns no reason and the checkbox comes up enabled; the user ticks it, and this method then falls through the `is None` guard and constructs the overlay, which raises `ValueError("a label marker overlay needs manual labels, predictions, or both")` straight out of `MenuHandlers.export_overlay_video` - there is no `try/except` anywhere on that path.

Guarding on emptiness here (`if not manual_labels and not predicted_labels:`), or having the `_export_*` helpers normalize `[]` to `(None, reason)`, closes it.

src/jabs/video_export/label_marker_overlay.py
gbeane, line 97: Same `None`-vs-empty split as in `CentralWidget.label_marker_overlay`, but internal to the class: `_caption()` and `_no_value_label()` are fed `manual_labels is not None` / `predicted_labels is not None`, while `marker_count` and `marker_values()` are driven by the emptiness-filtered `sources` property.

A source that is present but empty therefore burns in `"Grooming labels (left) and predictions (raw, right)"` and the `"no label or prediction"` legend entry while only one marker is ever drawn - so the caption names the wrong square. Deriving the caption/legend flags from the same emptiness test the drawing uses would keep them in step by construction, which is what the class docstring promises the constructors do.

gbeane (commented):
Correctness review of the label/prediction overlay change. The refactor is coherent: the geometry math in the player and the exporter agree (`group_width = n*size + (n-1)*pair_gap`, `group_x = x - group_width - gap`, `stride = size + pair_gap`), the rename is complete with no stale `PredictionOverlay` / `prediction_overlay=` references, and all three `native_label_marker_sizes` callers were updated for the new 3-tuple. `tests/ui tests/video_export tests/overlay_drawing` pass (445).

Five findings inline, one medium and four low. The medium one is a `None`-vs-empty-sequence mismatch between `label_marker_overlay()`'s guard and `LabelMarkerOverlay.__post_init__`'s validation.

Things I checked and found fine: the `_label_overlay_populated` flag is only ever set `True` alongside a real push and `False` alongside a real clear, so I could not construct a stale-marker case from it; `_showing_postprocessed_predictions` is safe on the new labels-only export path because both prediction dicts default to `{}`; and the multi-class LUT / `class_names` staleness check is unchanged from `main`.

src/jabs/ui/main_window/menu_handlers.py
gbeane, line 229: Performance regression worth a look: `label_markers_unavailable()` here and `central_widget.label_marker_overlay(...)` after the dialog both go through `_export_manual_labels()` / `_export_predicted_labels()`, and those do the *full* work rather than a cheap availability check - `_build_multiclass_overlay_labels()` allocates `num_identities x num_frames` arrays, `build_multiclass_label_array` walks every behavior, and `_get_prediction_list()` rebuilds the whole prediction list.

Before this change the overlay was built exactly once. Now a multi-class export does two full builds of both sources, and the first one runs synchronously *before* the options dialog is even shown, so on a long video the UI stalls on a menu click with nothing on screen to explain it. Caching the result of the availability pass and reusing it after the dialog would restore the old cost while keeping the "caption depends on the choice" ordering.

src/jabs/ui/player_widget/overlays/label_overlay.py
gbeane, line 102: The new loop added an `identity >= len(values)` guard but no frame guard, while its export counterpart `LabelMarkerOverlay._value` bounds-checks *both* identity and `frame_index` and documents why ("Labels and predictions can be shorter than the video").

In `BOTH` mode the two sources can have different per-identity lengths - manual labels are sized to `VideoLabels` / pose frames, predictions come from the saved prediction record - and the codebase already acknowledges video-vs-pose frame-count mismatch via `_frame_count_mismatch_message`. When `current_frame` exceeds one source's length this raises `IndexError` inside `paintEvent`.

Pre-existing for the single-source modes, but `BOTH` doubles the chance of hitting a short array, and the asymmetry with the exporter is worth closing here.

src/jabs/ui/main_window/central_widget.py
gbeane, line 1560: This claim (and the matching line in the PR description) overstates what the short-circuit does: `_label_button_common` still calls `self._player_widget.reload_frame()` unconditionally, immediately before `self._set_label_track()`.

So with the overlay switched off an edit goes from two frame reloads (the explicit one plus the one inside `set_labels`) down to one, not to zero - the displayed frame is still reloaded on every edit. Worth rewording to "does not reload the displayed frame a second time", or dropping the explicit `reload_frame()` in `_label_button_common` if the intent really was zero.

PR #479, opened 2026-09-18
Validate DANDI-required subject metadata before writing NWB

gbeane:
Resolves [KLAUS-706](https://jacksonlaboratory.atlassian.net/browse/KLAUS-706).

`jabs-cli convert-to-nwb` silently produced NWB files that the EMBER-DANDI archive rejects. The archive runs `nwbinspector` and blocks upload on any `CRITICAL` finding, and the default invocation produced three of them per file:

```
[CRITICAL] check_subject_age: Subject is missing age and date_of_birth.
[CRITICAL] check_subject_sex: Subject.sex is missing.
[CRITICAL] check_subject_species_exists: Subject species is missing.
```

Species, sex and age are biological facts that cannot be derived from a pose file, so they have to come from `--subjects` — which was documented as optional.

## What changed

Subject metadata is now validated **before anything is written**, failing with a per-identity breakdown:

```
Error: Subject metadata required by the DANDI archive is missing or malformed:
  subject_1: species is missing; sex is missing; age or date_of_birth is missing
  subject_2: species 'mouse' must be Latin binomial (e.g. 'Mus musculus') or an NCBI taxonomy IRI
  subject_3: sex 'male' must be one of 'M', 'F', 'O', 'U'
  subject_4: weight '25g' must be '[numeric] [unit]' with a space, e.g. '25 g' or '0.025 kg'
Every identity needs subject_id, species, sex, and age or date_of_birth. Supply them with
--subjects; see docs/user-guide/nwb-export.md (Subjects JSON format).
```

Validating up front matters because per-identity output writes one file per identity in a loop — failing partway would leave an incomplete, unpublishable set on disk.

An unmatched `--subjects` key is a warning rather than an error, since it fails anyway via the identity it leaves empty; the warning is what explains why:

```
WARNING These --subjects keys match no identity and were ignored: subject_0.
        Valid identity names for this pose file are: subject_1, subject_2, subject_3, subject_4
```

## Structure

- **`src/jabs/scripts/cli/dandi_subject_metadata.py`** (new) — the rules. `subject_metadata_problems()` is pure and per-identity; `validate_subjects()` resolves against a `PoseData` and reports every problem across every identity at once. DANDI policy lives here rather than in `jabs-io`, which stays a general-purpose IO layer.
- **`resolve_identity_subjects()`** (new, `jabs-io`) — shares the writer's name-sanitizing and metadata-lookup logic so validation sees exactly what the writer *will* resolve, instead of reimplementing it and drifting out of sync.

The rules were derived by running the real validator rather than read off its docs, and the three regexes are copied from `nwbinspector` verbatim so the two agree. That surfaced four things worth knowing:

1. The DANDI config elevates **more** checks than the three failures above — `check_subject_species_form`, `check_subject_id_no_slashes` and `check_subject_proper_age_range` as well. Presence checks alone would not have been enough.
2. `age` accepts **ranges** (`P1D/P3D`, `P90Y/`), so a plain ISO-duration check would have rejected valid input.
3. `sex` is **species-dependent** — C. elegans uses `XO`/`XX`, not `M`/`F`/`O`/`U`.
4. `species` requires a lowercase epithet, so `"Mus Musculus"` fails while `"Mus musculus"` passes.

## Two bugs found along the way

- **`subject_id` is auto-defaulted by the writer**, so requiring it in the metadata dict would have been a false positive. Validating the *effective* value instead means an unsanitized external ID such as `mouse/a` — which becomes the `subject_id` verbatim and trips `check_subject_id_no_slashes` — is now caught locally rather than at the archive.
- **The docs recommended `weight: "25g"`**, which fails a CRITICAL check; the accepted form needs a space (`"25 g"`). Fixed in both doc copies.

Also corrected the CLI help for `--subjects`, which documented the fallback keys as `subject_0, subject_1, …` while the writer generates `subject_1, subject_2, …` — following the help exactly matched zero identities.

## Verification

- Default invocation: fails with **0 files written**.
- With a valid subjects file: all four per-identity outputs report **zero CRITICAL** findings under `nwbinspector --config dandi` (the 5 remaining messages are session-level `BEST_PRACTICE_SUGGESTION`s, cleared via `--session-metadata`, and none block upload).
- 93 new tests. Full suite green: 1178 root, 352 jabs-io, 104 jabs-core, 113 jabs-behavior. Lint and format clean.

The `resolve_identity_subjects` tests are deliberately **not** `importorskip`-guarded — the function touches no NWB types, and this was verified with `_NWB_AVAILABLE = False`, so a plain `uv sync` environment still covers it.

## Notes for review

- **Behavior change:** conversions that previously succeeded without `--subjects` now fail. This was the explicitly chosen option over a `--dandi`/`--strict` opt-in flag, on the grounds that silently emitting unpublishable files is worse. Docs in both copies now describe `--subjects` as required.
- `nwbinspector` was **not** added as a dependency, so no test asserts zero CRITICAL findings directly — that check was manual, via an ephemeral environment. `uv.lock` is untouched. Adding it to a test group is a reasonable follow-up if we want that assertion in CI.
- Validation also runs for `--multisubject`, which is correct but does not make that mode archivable — its output cannot be read by `nwbinspector` at all. That is [KLAUS-707](https://jacksonlaboratory.atlassian.net/browse/KLAUS-707).

[KLAUS-706]: https://jacksonlaboratory.atlassian.net/browse/KLAUS-706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KLAUS-707]: https://jacksonlaboratory.atlassian.net/browse/KLAUS-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        Human-readable problem descriptions, one per failed requirement, ordered
+        so that missing required fields come before malformed optional ones.
+        Empty when the metadata satisfies every requirement checked here.
+    """
+    problems: list[str] = []
+
+    # The writer defaults subject_id rather than leaving it unset, so validate the
+    # value that will actually be written.
+    subject_id = metadata.get("subject_id") or default_subject_id
```
gbeane: **medium** - the validator falls back to the default on any *falsy* `subject_id`, but the writer falls back only on an *absent key*.

`metadata.get("subject_id") or default_subject_id` substitutes the default for `""` and `None`. The writer uses `raw_meta.get("subject_id", raw_key)` (`nwb.py:405`), which only substitutes when the key is missing:

- `{"subject_id": ""}` validates clean, and `_make_subject` produces `Subject(subject_id='')`.
- `{"subject_id": null}` validates clean, `_make_subject` filters the `None` out, and the Subject is written with no `subject_id` at all.

Both trip `check_subject_id_exists`, the CRITICAL finding this pre-flight check exists to catch. Verified directly: `_make_subject({"subject_id": ""})` -> `subject_id == ''`; `{"subject_id": None}` -> `None`.

A present-but-blank or `None` `subject_id` should be reported as a problem rather than silently replaced by the default.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+    age_missing = _missing(metadata, "age")
+    dob_missing = _missing(metadata, "date_of_birth")
+    if age_missing and dob_missing:
+        problems.append("age or date_of_birth is missing")
+    if not age_missing and not _is_iso_duration(str(metadata["age"])):
+        problems.append(
+            f"age {metadata['age']!r} must be an ISO 8601 duration (e.g. 'P70D', 'P2Y') "
+            "or a range (e.g. 'P1D/P3D', 'P90Y/')"
+        )
+    if not dob_missing:
```
gbeane, line 224: **medium** - blank `date_of_birth` is skipped by the validator but not by the writer, and it crashes mid-write.

`_missing()` classifies `date_of_birth: ""` (or whitespace-only) as missing, so `dob_missing` is true and this whole block is skipped - no problem reported. The writer does not skip it: `_make_subject` runs `datetime.fromisoformat(dob_raw)` for any non-`None` value.

Concretely, `{"species": "Mus musculus", "sex": "M", "age": "P70D", "date_of_birth": ""}` passes validation, then the write raises `ValueError: Invalid isoformat string: ''`. In per-identity mode that happens *inside* the write loop, so earlier identities' files are already on disk - the partial, unpublishable set that validating up front was meant to prevent.

This is the one I'd fix first. Verified both halves.

gbeane (commented):
## Review: DANDI subject-metadata pre-flight

The design here is right - a pure per-identity rule function plus a resolver shared with the writer, run before anything is written. The findings below are all about the *boundary* between the two halves, not the structure.

### Shared root cause (findings 1-4)

Findings 1 through 4 are one mismatch showing up in four places: **the validator treats any falsy value as absent** - substituting a default (`subject_id`) or skipping the form check entirely (`age`, `date_of_birth`) - while **the writer only filters `None` or an absent key**. So `""`, and in one case a present `null`, sails through pre-flight and then either crashes the write or reaches the archive as the exact CRITICAL finding this PR exists to prevent.

Worth aligning both sides on a single "is this value usable" helper (or having `_make_subject` drop blanks the same way `_missing` does), so the next field added to `_SUBJECT_FIELDS` doesn't reintroduce it field-by-field.

**Prioritize #2.** It is the only one that raises *inside* the per-identity write loop, so it leaves the partial, unpublishable file set that validating up front was meant to rule out.

All seven were confirmed by executing `subject_metadata_problems` and `PoseNWBAdapter._make_subject` against the same inputs.

### Also, cosmetic

`_WEIGHT_FORM` accepts `μg`, but the docs unit list reads "Units: kg, g, mg, ug, ng, pg" - the micro-sign variant is missing from the table in both doc copies. Not worth an inline comment.

### Verified as fine

- Both doc copies are byte-identical after the change, and no stale `"25g"` recommendation remains anywhere in `docs/` or `src/jabs/resources/docs/`.
- `run_conversion` has exactly one non-test caller (`cli.py:627`), so the new `ValueError` has no other callers to break.
- `resolve_identity_subjects` genuinely touches no NWB types, and `nwb.py`'s existing import guards mean the un-`importorskip`-ed test file is correct under a plain `uv sync`.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+        # catch an unparseable value here rather than mid-write.
+        if isinstance(dob, str):
+            try:
+                datetime.datetime.fromisoformat(dob)
+            except ValueError:
+                problems.append(
+                    f"date_of_birth {dob!r} is not an ISO 8601 datetime "
+                    "(e.g. '2024-01-15T00:00:00+00:00')"
+                )
+        elif not isinstance(dob, datetime.datetime):
```
gbeane, line 236: **medium** - the validator accepts a `datetime` for `date_of_birth`, but the writer cannot handle one.

This branch deliberately allows `datetime.datetime`, and `test_datetime_date_of_birth_is_accepted` locks that in. `_make_subject` calls `datetime.datetime.fromisoformat(dob_raw)` unconditionally on any non-`None` value, which raises `TypeError: fromisoformat: argument must be str` for a datetime object.

So a programmatic caller of `run_conversion` passing an already-parsed datetime passes pre-flight and then dies during the write. Verified: `_make_subject({"date_of_birth": datetime.datetime(2024, 1, 1)})` -> `TypeError`.

Either reject non-`str` here, or let `_make_subject` pass a datetime straight through.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+
+    sex = metadata.get("sex")
+    if _missing(metadata, "sex"):
+        problems.append("sex is missing")
+    else:
+        allowed = _C_ELEGANS_SEX_VALUES if species in _C_ELEGANS_SPECIES else _SEX_VALUES
+        if str(sex) not in allowed:
+            problems.append(f"sex {sex!r} must be one of {', '.join(repr(v) for v in allowed)}")
+
+    age_missing = _missing(metadata, "age")
```
gbeane, line 202: **medium** - same present-but-blank hole on the `age` branch.

`age_missing` short-circuits the duration form check on line 121, so `age: ""` alongside a valid `date_of_birth` reports nothing. The writer only filters `None`, so `Subject(age="")` is written. `check_subject_age` sees a non-`None` value that fails the duration regex and flags it CRITICAL - the file is rejected at the archive even though pre-flight was clean.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+    Mirrors ``check_subject_age``, which accepts a plain duration (``"P70D"``) as
+    well as a ``/``-separated range with either bound optionally blank
+    (``"P1D/P3D"``, ``"P90Y/"``).
+    """
+    if _DURATION.fullmatch(value):
+        return True
+    if "/" not in value:
+        return False
+    lower, _, upper = value.partition("/")
+    return all(bound == "" or bool(_DURATION.fullmatch(bound)) for bound in (lower, upper))
```
gbeane, line 75: **low** - this allows an empty *lower* range bound, which `nwbinspector` does not.

`all(bound == "" or ...)` means both bounds may be blank, so `age: "/"` and `age: "/P3D"` are accepted. `check_subject_age` requires the lower bound to be a valid duration and permits only the *upper* bound to be empty - and the docstring above plus the docs table only claim `"P90Y/"` (open upper) as valid.

Both forms therefore pass pre-flight and fail at the archive. Verified: `subject_metadata_problems({..., "age": "/"})` returns `[]`.

Requiring a non-empty, well-formed lower bound would close it.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+                ", ".join(unused),
+                ", ".join(entry.identity_name for entry in resolved),
+            )
+
+    failures = [
+        (entry.identity_name, problems)
+        for entry in resolved
+        if (
+            problems := subject_metadata_problems(
+                entry.metadata, default_subject_id=entry.lookup_keys[0]
```
gbeane, line 301: **low** - a non-object per-identity value produces an opaque `AttributeError` instead of a validation error.

The CLI only checks that the subjects JSON is a dict at the *top* level (`cli.py:607`), so `{"subject_1": "Mus musculus"}` reaches `subject_metadata_problems`, which calls `metadata.get(...)` and raises `AttributeError: 'str' object has no attribute 'get'`. The user sees `Error: 'str' object has no attribute 'get'`.

The crash pre-exists in the writer (`{**raw_meta, ...}` on a str), but validation is now the first thing to touch the value and is the natural place to say `subject_1: metadata must be a JSON object, got string`.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Raises:
+        ValueError: If any identity's subject metadata is missing or malformed,
+            listing the offending identities and their problems.
+    """
+    resolved = resolve_identity_subjects(data)
+
+    supplied = set((data.subjects or {}).keys())
+    if supplied:
+        matched = {key for entry in resolved for key in entry.lookup_keys}
```
gbeane: **low** - the unused-key warning misses a key that is shadowed rather than unmatched.

`matched` is the union of *all* `lookup_keys`, but when sanitizing changes a name both the raw and the sanitized key land in `lookup_keys` while `_resolve_subject` only ever uses the raw one.

With `external_ids = ["mouse/a"]` and `subjects = {"mouse/a": {...}, "mouse_a": {...}}` - the exact shape `test_raw_key_wins_over_the_sanitized_name` asserts - the `"mouse_a"` entry is silently discarded and no warning is logged. That is the case the warning was added to explain.

gbeane:
Thanks both — pushed `6fc11ab9`. Every claim was re-checked by running `subject_metadata_problems` and `PoseNWBAdapter._make_subject` on the same input, and the archive consequences against `nwbinspector` under its `dandi` config. Seven findings confirmed and fixed; three did not reproduce and are explained below.

## Fixed at the root

The review's framing was right: **the validator treated any falsy value as absent while the writer only filtered `None`.** Rather than patch that field by field, both halves now share one predicate, `subject_value_is_absent()` in `jabs-io` — `None` or a blank/whitespace string means "not supplied", for the validator and the writer alike. A new field added to `_SUBJECT_FIELDS` therefore cannot reintroduce the mismatch.

That closes four findings at once:

| Input | Before | Now |
|---|---|---|
| `date_of_birth: ""` | `ValueError` inside the write loop → partial file set | omitted by both |
| `age: ""` | `Subject(age="")` → `check_subject_age` CRITICAL | omitted by both |
| `subject_id: null` | Subject written with no id → `check_subject_id_exists` CRITICAL | falls back to the identity name |
| `subject_id: ""` | replaced by the default in validation only | falls back in both |

Verified end to end: a subjects file with blank `subject_id` **and** blank `date_of_birth` — which previously raised mid-write — now converts and reports **zero CRITICAL** findings on every output file.

`date_of_birth` as a `datetime` is now handled by `_make_subject` rather than rejected by the validator; a naive one gets UTC, matching the string path. The old test that locked in the broken expectation is replaced with one asserting the value survives the write.

## Also fixed

- **Reversed age ranges** (`P3D/P1D`) are now rejected. Good catch — `check_subject_proper_age_range` is elevated to CRITICAL by the DANDI config and I had only validated bound *syntax*. The implementation is deliberately conservative: each bound converts to a **range** of possible day-counts (a month is 28–31 days), and a range is only rejected when the bounds provably cannot overlap. So `P3D/P1D` fails while calendar-ambiguous `P1M/P30D` is left to the archive's exact arithmetic rather than being wrongly blocked here.
- **Non-object identity metadata** (`{"subject_1": []}`) reports `metadata must be a JSON object, got list` instead of raising `AttributeError`.
- **Shadowed `--subjects` keys** are now warned about. `IdentitySubject` gained `matched_key` — the one key the writer actually reads — so the warning uses that rather than the union of candidates, and a key silently losing to a higher-precedence one is reported.
- **CLI docstring examples** now all pass `--subjects`, and the option help no longer claims `subject_id` is required when it is defaulted.
- **The no-partial-output guarantee is now a test**, asserting `save` is not called and no `.nwb` lands on disk.
- Docs: `μg` added to the unit list, blank-means-absent documented, and the age row now states both bounds may be open and that ranges must increase.

## Three that did not reproduce

- **"Empty lower bound fails at the archive"** (`age="/"`, `"/P3D"`). Verified **CLEAN**. `check_subject_age` accepts a blank bound on *either* side — `lower_valid = fullmatch(...) or lower == ""` — so the implementation already matched it. Only my docstring and docs table were narrower than the code; those are now corrected rather than the code.
- **`subject_id: ""` trips `check_subject_id_exists`.** Verified **CLEAN** — that check tests for `None`, not emptiness. The validator/writer mismatch was real and is fixed, but the archive consequence was not; the `null` case is the one that actually failed.
- **"Numeric weight bypasses the check."** A float weight is **CLEAN** at the archive, so it is not an archive risk and is now explicitly accepted. There is a real bug nearby though, so thank you for the pointer: pynwb's `Subject` takes only `str` or `float`, so an **int** from JSON (`{"weight": 25}`) raises `TypeError` during the write. That is now reported by the validator instead.

Full suite green: 1454 root, 355 jabs-io, 104 jabs-core, 113 jabs-behavior; lint and format clean.

src/jabs/scripts/cli/dandi_subject_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+            sanitized container name. Checked in place of an absent ``subject_id``
+            so this agrees with what is actually written, and so an unsanitized
+            external ID containing ``/`` is still caught.
+
+    Returns:
+        Human-readable problem descriptions, one per failed requirement, ordered
+        so that missing required fields come before malformed optional ones.
+        Empty when the metadata satisfies every requirement checked here.
+    """
+    problems: list[str] = []
```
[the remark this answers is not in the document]
gbeane: Fixed in 6fc11ab9: `subject_metadata_problems` now type-guards and reports `metadata must be a JSON object, got list` instead of raising.

src/jabs/scripts/cli/cli.py
```diff
@@ -526,12 +526,15 @@ def cross_validation(
     default=None,
     help=(
         "Path to a JSON file containing per-animal biological metadata. "
+        "Required: the DANDI archive rejects output without it. "
```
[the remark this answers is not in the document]
gbeane: Fixed in 6fc11ab9: every docstring example now passes `--subjects`, with a line above them saying it is required.

src/jabs/scripts/cli/convert_to_nwb.py
```diff
@@ -274,6 +278,11 @@ def run_conversion(
 
     pose_data = pose_to_pose_data(pose, subjects=subjects)
 
+    # Validate before writing: per-identity output writes one file per identity in a
+    # loop, so failing partway would leave an incomplete set on disk, and the whole
+    # set would be unpublishable anyway.
+    validate_subjects(pose_data)
```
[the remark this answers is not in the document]
gbeane: Fixed in 6fc11ab9: `test_run_conversion_rejects_invalid_subjects_before_saving` asserts `save` is not called and no `.nwb` lands on disk.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Fixed in 6fc11ab9: both halves now share `subject_value_is_absent()`, so validator and writer fall back on exactly the same values.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Fixed in 6fc11ab9: `_make_subject` omits blank values, so this no longer raises mid-write. Verified end to end - a blank `date_of_birth` now converts with zero CRITICAL findings.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Fixed in 6fc11ab9: `_make_subject` now accepts a `datetime` directly (naive values get UTC), and the test asserts it survives the write rather than only passing validation.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Fixed in 6fc11ab9: a blank `age` is omitted by the writer too, so `Subject(age="")` is no longer written.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Fixed in 6fc11ab9: reported as `metadata must be a JSON object, got <type>` rather than an `AttributeError`.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Fixed in 6fc11ab9: `IdentitySubject` gained `matched_key` - the one key the writer actually reads - and the warning now uses that, so a shadowed key is reported.

src/jabs/scripts/cli/dandi_subject_metadata.py
gbeane: Checked this against both layers of the archive's pipeline, and the premise doesn't hold — but there was a real problem underneath it, now fixed in `221af544`.

**`nwbinspector` accepts a blank lower bound.** `check_subject_age` validates each side independently and permits either to be empty:

```python
lower_valid = re.fullmatch(duration_regex, lower) or lower == ""
upper_valid = re.fullmatch(duration_regex, upper) or upper == ""
if lower_valid and upper_valid:
    return None
```

**DANDI's own `parse_age` accepts them too**, so this isn't nwbinspector being laxer than the archive:

```
parse_age('P90Y/')  -> ('P90Y/', 'Birth')
parse_age('/P3D')   -> ('/P3D', 'Birth')
parse_age('/')      -> ('/', 'Birth')
parse_age('P3D/P1D')-> ValueError: The upper limit has to be larger than the lower limit
```

So `"/P3D"` is legitimately valid — "no older than 3 days" is a meaningful statement — and requiring a well-formed lower bound would have rejected input the archive is happy with. I've left that case accepted and instead widened the docstring and the docs table, which were the things actually claiming only `"P90Y/"` was allowed.

**The real problem was `"/"`.** You were right that something was wrong here, just not that the archive would catch it — it won't. A bare `"/"` states no age whatsoever, yet it satisfies our "age or date_of_birth is missing" requirement, so it was a way to pass the age check with a value carrying zero information. Since this pre-flight exists to guarantee *publishable* metadata and not merely *accepted* metadata, being stricter than both validators is right here. `"/"` is now rejected:

```
age '/' gives neither bound, so it states no age at all; supply a duration
(e.g. 'P70D') or a range with at least one bound (e.g. '/P3D', 'P90Y/')
```

One open bound stays valid. Test coverage distinguishes the three cases.

Incidentally, that `parse_age` output also raises the stakes on the reversed-range fix from `6fc11ab9`: `P3D/P1D` isn't just an nwbinspector finding, it hard-errors DANDI's metadata extraction.

Leaving this thread for you to close, since I've pushed back on the mechanism rather than simply applying it.

